PR #123, opened 2023-04-19
Replace express with native http server

TrySound:
Here dropped express in favour of node native http server. Express is basically same thing with polluted request and response.

For static serving in the future can be used https://github.com/lukeed/sirv

```diff
du -ckd1 packages/partykit/dist
- 5512
+ 2980
```

Also see that nice red wall in lockfile

### DIFF
--- a/.changeset/flat-walls-mate.md
+++ b/.changeset/flat-walls-mate.md
@@ -1,0 +1,5 @@
+---
+"partykit": patch
+---
+
+replace express with node http server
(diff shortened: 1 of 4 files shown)


threepointone:
Love it. Could you rebase and add a changeset? Happy to land right after.

TrySound:
Alright, done

threepointone (approved):
Thank you very much!

PR #126, opened 2023-04-24
chore: enable typescript-eslint/recommended-requiring-type-checking

### LINKED ISSUE #125: Using recommended-requiring-type-checking from typescript-eslint
👋 congrats on open sourcing!

I see you're already using a bunch of rules from typescript-eslint: https://github.com/partykit/partykit/blob/baa74442a567b303ef0f93edcf3baa6622e0a749/.eslintrc.js. Great!

We actually have more rules available: https://typescript-eslint.io/linting/configs#recommended-requiring-type-checking. Right now, you're using `plugin:@typescript-eslint/recommended` but not ` plugin:@typescript-eslint/recommended-requiring-type-checking`. The latter enables quite a few useful rules -including `no-floating-promises`. Would you be up for me sending a PR enabling more for you? 😄

JoshuaKGoldberg:
Fixes #125.

I avoided enabling any lint rule that had multiple existing failures. The `MessageEvent` type has a lot of implicit `any`s that tripped them up. Otherwise, changes are commented inline.

Recognizing that this is your repository and I have little to no context on it, please consider everything in this PR as a suggestion - I'm happy to revert any changes you're not a fan of! ❤️ 

[hyperfine](https://github.com/sharkdp/hyperfine) measurements on this PR:

```plaintext
  Time (mean ± σ):      3.006 s ±  0.083 s    [User: 5.989 s, System: 0.369 s]
  Range (min … max):    2.907 s …  3.103 s    10 runs
```

...compared to `main` (without `--cache`):

```plaintext
  Time (mean ± σ):      2.933 s ±  0.035 s    [User: 5.889 s, System: 0.371 s]
  Range (min … max):    2.873 s …  3.006 s    10 runs
```


packages/partykit/src/bin.ts
```diff
@@ -252,6 +252,7 @@ The above copyright notice and this permission notice shall be included in all c
 THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
+// eslint-disable-next-line @typescript-eslint/unbound-method
 const fn = new Intl.Collator(undefined, { numeric: true }).compare;
```
JoshuaKGoldberg, line 256: Technically, `unbound-method` is correct on this case. But it seems to work? Maybe the `.d.ts` types for `Collator` are incorrect? Since this is under an MIT license I was unsure whether it's best to change this line or just ignore the rule once?

https://typescript-eslint.io/rules/unbound-method

(this comment is in part serving as user research for me - do those docs make sense in this context?)

packages/y-partykit/src/storage.ts
```diff
@@ -51,10 +51,10 @@ export async function levelGet(
 ): Promise<Uint8Array | null> {
   const prefix = keyEncoding.encode(key);
 
-  const res = (await db.list({
+  const res = await db.list<Uint8Array>({
     start: prefix,
     end: `${prefix}#zzzzz`,
-  })) as Map<string, Uint8Array>;
+  });
```
JoshuaKGoldberg, line 57: https://github.com/typescript-eslint/typescript-eslint/issues/6951 😄 - love to file bugs when adopting the library in the wild!

I'm not sure that this change is net positive. Previously, the code was using a technically unsafe `as`. Now the code is using a technically unsafe explicit type parameter (https://effectivetypescript.com/2020/08/12/generics-golden-rule/#:~:text=Rule%3A%20If%20a%20type%20parameter,if%20you%20actually%20need%20it.). 🤔

packages/y-partykit/src/storage.ts
```diff
@@ -97,13 +97,15 @@ export async function levelPut(
 }
 
 function groupBy<T>(arr: T[], fn: (el: T) => string): Map<string, T[]> {
-  const map = new Map();
+  const map = new Map<string, T[]>();
```
JoshuaKGoldberg, line 100: Previously, this was implicitly `Map<any, any>`. TypeScript can't infer the type of `map` when there aren't any initial values.

https://typescript-eslint.io/rules/no-unsafe-return

packages/y-partykit/src/storage.ts
```diff
@@ -97,13 +97,15 @@ export async function levelPut(
...
     const key = fn(el);
-    if (!map.has(key)) {
+    const existing = map.get(key);
+    if (existing) {
+      existing.push(el);
+    } else {
       map.set(key, []);
     }
-    map.get(key).push(el);
   }
```
JoshuaKGoldberg, line 109: This is a common gripe with `map.get`/`.set` in TypeScript: the fact that `.has` returned `true` doesn't cause TypeScript to narrow the type of `map`. In other words, `map.has(key)` doesn't tell TypeScript that `map.get(key)` will later return a non-`undefined` value. 

I generally work around this with a `const existing = ...`. But an alternate change could be to write `map.get(key)!` instead.

package.json
```diff
@@ -11,7 +11,7 @@
     "start": "npm start -w partykit",
     "build": "npm run build -w partykit -w y-partykit -w partysocket -w partykit.io",
     "check": "npm run lint && npm run repocheck && npm run typecheck && npx vitest --watch=false",
-    "lint": "npx eslint \"**/*.[tj]s?(x)\" --max-warnings=0 --cache",
+    "lint": "npx eslint \"**/*.[tj]s?(x)\" --max-warnings=0 --report-unused-disable-directives",
```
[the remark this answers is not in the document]
threepointone: noted, TIL

packages/partykit/src/bin.ts
threepointone: I just copied this from elsewhere so I dunno  what the right move is here. happy to have the disable for now, we can revisit if it causes issues

packages/y-partykit/src/storage.ts
threepointone: huh TIL

threepointone (approved):
Thanks so much for this, very happy to see the codebase get healthier. I appreciate you!

threepointone:
kinda odd that tests are failing, there haven't been any functional changes

threepointone:
Aight, let's land this. Thank you very much!

PR #128, opened 2023-04-28
Don't crash dev server on unhandled exceptions

### LINKED ISSUE #116: dev: error shouldn't crash the server
right now an error can crash the whole server, we should be able to guard against that.

philipp-spiess:
Closes #116

Since we run the userspace code in the same node process that runs the `dev` command, any unhandled exception crashes that process. 

I’m not sure what the ideal behavior is. Maybe we should fork the websocket server that handles userspace code into a seperate processes and restart it so that all server-side state is reset? We don't really know _what_ causes the crash and it doesn't have to be related to a specific connection (i.e the combination of multiple connections could cause some invalid state).

For now, disconnecting all clients seems like a reasonable dev behavior. This should reset most of the state and let you know that something went wrong even if you're not looking at the console.

### DIFF
--- a/.changeset/wise-sheep-beam.md
+++ b/.changeset/wise-sheep-beam.md
@@ -1,0 +1,5 @@
+---
+"partykit": patch
+---
+
+Don't crash dev server on unhandled exceptions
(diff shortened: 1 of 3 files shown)


threepointone:
Thanks for sending this way back @philipp-spiess. The whole project's changed since then, I don't think we crash any more in these circumstances. Thank you!

PR #133, opened 2023-05-08
Make y-partykit compatible with both esm and cjs

mellson:
Similar to [this PR](https://github.com/partykit/partykit/pull/69) for partysocket, I need y-partykit working as both esm and cjs.

### DIFF
--- a/packages/y-partykit/package.json
+++ b/packages/y-partykit/package.json
@@ -2,18 +2,27 @@
   "name": "y-partykit",
   "version": "0.0.0",
   "description": "y.js on partykit!",
-  "main": "dist/index.js",
+  "main": "dist/cjs/index.js",
+  "module": "dist/esm/index.js",
   "exports": {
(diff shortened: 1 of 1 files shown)


packages/y-partykit/package.json
```diff
@@ -22,8 +31,7 @@
     "@types/lodash.debounce": "^4.0.7"
   },
   "files": [
-    "dist/*.js",
-    "dist/*.js.map",
```
mellson, line 26: This PR removes the source maps from the published package. I did this to mirror partysocket, but I can add it for both if preferred?

threepointone: feel free to add them back in a followup PR, lemme land this and unblock you right now

threepointone (commented):
Thank you very much! Note to self: We should start looking at a build framework to automate all this in the future.

threepointone (approved):

PR #134, opened 2023-05-09
Use `.mjs` for ESM files and add `types` condition to `exports`

Andarist:
This is kinda a follow-up to https://github.com/partykit/partykit/pull/133

cc @mellson @threepointone

### DIFF
--- a/packages/partysocket/package.json
+++ b/packages/partysocket/package.json
@@ -6,15 +6,18 @@
   "module": "dist/esm/index.js",
   "exports": {
     ".": {
-      "import": "./dist/esm/index.js",
+      "types": "./index.d.ts",
+      "import": "./dist/esm/index.mjs",
       "require": "./dist/cjs/index.js"
     },
     "./ws": {
-      "import": "./dist/esm/ws.js",
+      "types": "./ws.d.ts",
+      "import": "./dist/esm/ws.mjs",
       "require": "./dist/cjs/ws.js"
     },
     "./react": {
-      "import": "./dist/esm/react.js",
+      "types": "./react.d.ts",
+      "import": "./dist/esm/react.mjs",
       "require": "./dist/cjs/react.js"
     }
   },
@@ -22,7 +25,7 @@
     "test": "echo \"Error: no test specified\" && exit 1",
     "clean": "rm -rf dist && rm -rf *.d.ts*",
     "build:cjs": "npx esbuild src/index.ts src/react.ts src/ws.ts --format=cjs --outdir=dist/cjs",
-    "build:esm": "npx esbuild src/index.ts src/react.ts src/ws.ts --format=esm --outdir=dist/esm",
+    "build:esm": "npx esbuild src/index.ts src/react.ts src/ws.ts --format=esm --outdir=dist/esm --out-extension:.js=.mjs",
     "build": "npm run clean && npm run build:cjs && npm run build:esm && tsc --project tsconfig.extract.json && mv dist/*.d.ts* ."
   },
   "files": [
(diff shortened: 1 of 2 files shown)


packages/partysocket/package.json
```diff
@@ -6,23 +6,26 @@
...
+      "import": "./dist/esm/react.mjs",
       "require": "./dist/cjs/react.js"
     }
   },
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
     "clean": "rm -rf dist && rm -rf *.d.ts*",
     "build:cjs": "npx esbuild src/index.ts src/react.ts src/ws.ts --format=cjs --outdir=dist/cjs",
-    "build:esm": "npx esbuild src/index.ts src/react.ts src/ws.ts --format=esm --outdir=dist/esm",
+    "build:esm": "npx esbuild src/index.ts src/react.ts src/ws.ts --format=esm --outdir=dist/esm --out-extension:.js=.mjs",
```
Andarist, line 28: You can't use `.js` extension for ESM files if your package is CJS (and it's CJS by default).

Alternatively to this you could output this:
```json
// ./dist/esm/package.json
{ "type": "module" }
```

The module format for `.js` files is looked up in "the closest package.json". So by using extra one-line package.json files you can actually use `.js` extension for all of your files.

packages/y-partykit/package.json
```diff
@@ -6,21 +6,23 @@
...
+      "import": "./dist/esm/provider.mjs",
       "require": "./dist/cjs/provider.js"
     }
   },
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
     "clean": "rm -rf dist && rm -rf *.d.ts*",
     "build:server:cjs": "npx esbuild src/index.ts src/storage.ts --outdir=dist/cjs --format=cjs",
-    "build:server:esm": "npx esbuild src/index.ts src/storage.ts --outdir=dist/esm --format=esm",
+    "build:server:esm": "npx esbuild src/index.ts src/storage.ts --outdir=dist/esm --format=esm --out-extension:.js=.mjs",
```
Andarist, line 23: I wonder... why this package has both `build:server` and `build:provider` scripts? It's a little bit less usual setup and I wonder if it's fully correct here and if it will stay correct with future changes

packages/partysocket/package.json
```diff
@@ -6,23 +6,26 @@
   "module": "dist/esm/index.js",
   "exports": {
     ".": {
-      "import": "./dist/esm/index.js",
+      "types": "./index.d.ts",
```
Andarist, line 9: Without this TypeScript wouldn't be able to find the types with `moduleResolution: node16`.

Note that with this change those packages will be flagged as "masquerading as CJS". I believe it's a problem but for practical purposes it's often OK-ish and many people ignore it. It's still better than "can't resolve the types" that the current setup results with.

packages/partysocket/package.json
mellson: Nice, thanks for the info! I think the change to `.mjs` files would be fine. Can you update the `files` array to include those as well? Right now, they only include `.js` files under `dist`

packages/partysocket/package.json
mellson: Sweet. And thanks for the link to `arethetypeswrong` site; I didn't know that existed.

threepointone:
Incredible, blessed with an @Andarist PR. Thank you so much!

packages/partysocket/package.json
Andarist: good catch, thanks! fixed that

Andarist:
> Incredible, blessed with an @Andarist PR.

For more to come in the future 🍻

PR #142, opened 2023-05-29
fix running dev

threepointone:


packages/partykit/src/cli.ts
```diff
@@ -279,7 +279,9 @@ export async function dev(options: {
   let wasmModules: Record<string, WebAssembly.Module> = {};
 
   const workerFacade = fs.readFileSync(
-    path.join(__dirname, "../facade/generated.js"),
+    fileURLToPath(
+      path.join(path.dirname(import.meta.url), "../facade/generated.js")
+    ),
```
TrySound, line 284: Afaik fs functions accept file url

threepointone: oh fr? I tried and it gave me an error, but maybe I did something wrong. will try again, thx!

TrySound: This one is nice. Fs accepts URL too.
```
const __dirname = new URL('.', import.meta.url)
```

https://blog.logrocket.com/alternatives-dirname-node-js-es-modules/

PR #143, opened 2023-06-04
use miniflare/workerd

threepointone:
This changes the local dev experience to use cloudflare's js runtime workerd (via miniflare). This has a lot of implications:
- the dev runtime now matches closer to what's running in production
- this means we'll get platform features more quickly (as an example, we can now connectto postgres databases)
- we can iterate on our features more quickly, since we don't have to maintain 2 separate implementations
- fewer bugs between dev and prod

The tradeoffs -
- it's quite heavy, so initial install is going to be slow
- won't run in stackblitz-like environments
- it's a brand new runtime, so we can expect some bugs
- setting up types is kind of annoying
- setting up a test environment is annoying

---

TODO: 
- [x] fix the types story, I have way too many expect-errors floating around 
- [x] when running the lexical example, this shows up: `An event handler returned a promise that will be ignored. Event handlers should not have a return value and should not be async functions.`
- [ ] this warning when running eslint `WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.`


examples/basic/src/client.ts
```diff
@@ -26,6 +27,7 @@ setInterval(() => {
   partySocket.send(`latency:${id}`);
 }, 1000);
 
+// @ts-expect-error I dunno how to type this
```
sadn1ck, line 30: adding "DOM" to examples/basic/tsconfig.json > compilerOptions.lib array should fix this (and the other ones as well. [base templates from vite](https://github.com/vitejs/vite/blob/471fba2d1cf647a0bea747739d4d90def60e83ce/packages/create-vite/template-react-ts/tsconfig.json#L5) have it, for reference)

docs: https://www.typescriptlang.org/tsconfig#lib

examples/blocknote/src/client.tsx
```diff
@@ -34,6 +34,7 @@ function Editor() {
   return <BlockNoteView editor={editor} />;
 }
 
+// @ts-expect-error I dunno how to fix this type TODO fix this
 // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
 const root = createRoot(document.getElementById("root")!);
```
sadn1ck, line 39: same "DOM" to compilerOptions.lib fix + assertion as `HTMLElement` ([ref to vite template](https://github.com/vitejs/vite/blob/471fba2d1cf647a0bea747739d4d90def60e83ce/packages/create-vite/template-react-ts/src/main.tsx#L6))

examples/basic/src/client.ts
threepointone: I tried this, unfortunately it clashes with some types in workers-types (specifically, 'globals' like `Request`/`Response` that have slight differences in the server environment. Ideally I'd like to be able to specify different types for `client.ts` and `server.ts` without making them separate packages. Any ideas?

sadn1ck: I think we can just have separate `tsconfig` for the client and server each inside a folder without splitting it into different packages (by split I assume you meant basic-client + basic-server)
image

although can't speak for the maintainability further down the line haha

threepointone: I wouldn't mind doing this, but now I can't just call `tsc` in the root here to do a full typecheck (or can I?)

sadn1ck: hmm yup you're correct, we can't. that's a problem

threepointone:
Landing this after a quick swim, lfg.

PR #157, opened 2023-07-10
feat: optional hibernation api

threepointone:
This introduces cloudflare's new hibernation api for durable objects. (see https://developers.cloudflare.com/workers/runtime-apis/durable-objects/#hahahugoshortcode-s2-hbhb) It kicks in when you specify `onMessage(){}` in the export. This should let us scale to thousands of connections on a single object, while also not charging for idle objects. Very nice.

--- 

DO NOT MERGE until we publish the server side version of this as well.


threepointone:
github doesn't let me approve my own request, but assume I've done so and feel free to iterate and land this when you feel comfortable

PR #158, opened 2023-07-10
Fixes empty `env` in dev mode

jevakallio:
The migration to miniflare/workerd seems to have broken `room.env` in development mode. Instead of the `vars` sourced from partykit.json and `--var` flags, `room.env` was being defined as the facade's env (a `DurableObjectNamespace` map).

I've verified this approach works for both the main durable object and any other parties in a multiparty scenario.

I'm not 100% certain that this fix is the best way of achieving the desired outcome, feedback welcome!

### DIFF
--- a/packages/partykit/src/bin.tsx
+++ b/packages/partykit/src/bin.tsx
@@ -114,9 +114,6 @@ program
   .option("-n, --name <name>", "name of the project")
   .option("--preview [name]", "deploy to preview environment")
   .action(async (scriptPath, options) => {
-    if (options.withVars) {
-      console.warn("--with-vars is not yet implemented");
-    }
     await cli.deploy({
       main: scriptPath,
       name: options.name,
--- a/packages/partykit/facade/source.ts
+++ b/packages/partykit/facade/source.ts
@@ -39,10 +39,6 @@ let didWarnAboutMissingConnectionId = false;
 
 class PartyDurable {}
 
-// injected by esbuild in dev mode
-const envVars = __ENV_VARS__;
-declare const __ENV_VARS__: Record<string, unknown>;
-
 function createDurable(Worker: PartyKitServer) {
   if (Worker.onConnect && typeof Worker.onConnect !== "function") {
     throw new Error(".onConnect is not a function");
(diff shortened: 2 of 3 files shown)


packages/partykit/src/bin.tsx
```diff
@@ -114,9 +114,6 @@ program
   .option("-n, --name <name>", "name of the project")
   .option("--preview [name]", "deploy to preview environment")
   .action(async (scriptPath, options) => {
-    if (options.withVars) {
```
jevakallio, line 117: This warning is outdated and caused some confusion while debugging this, so I removed it. Tested that `--with-vars` is implemented and works as expected.

jevakallio: The initially reported issue with `partykit.json` variables not being defined in deployed package was an user error on my part, I wasn't aware of the `with-vars` semantics. I've verified that works correctly.

We may want to change the default behaviour there as discussed, but let's do that as a separate piece of work.

threepointone:
I *think* `server.onBundleUpdate()` accepts a `bindings` object (via https://miniflare.dev/get-started/api). Could you try `bindings: config.vars` over here?
https://github.com/partykit/partykit/blob/46d871d98a50362efaa85b5c6ff9bbbdf299aba1/packages/partykit/src/dev.tsx#L267-L269

packages/partykit/src/bin.tsx
threepointone: good catch, thank you!

threepointone (approved):
Awesome, thanks for fixing my goof! Sign the CLA, and feel free to merge!

jevakallio:
> I _think_ `server.onBundleUpdate()` accepts a `bindings` object (via https://miniflare.dev/get-started/api). Could you try `bindings: config.vars` over here?
> 
> https://github.com/partykit/partykit/blob/46d871d98a50362efaa85b5c6ff9bbbdf299aba1/packages/partykit/src/dev.tsx#L267-L269

Thank you, this works and reads a lot cleaner! Refactored to use `bindings`.

While testing this, I noticed that we're leaking the DO namespace stubs in `room.env`, and in addition the behaviour is slightly different in both dev and the deployed platform.

For, example, see see the behaviour for the following `partykit.json` config:
```json
{
  "main": "index.ts",
  "vars": {
    "a": "a"
  },
  "parties": {
    "other": "./other.ts"
  }
}
```

In `dev`, `room.env` is:
```
{
  "a":"a",
  "MAIN_DO":{...},
  "other":{...}
}
```

In a deployed worker, `room.env` is:
```
{
  "a":"a",
  "other":{...}
}
```

Whereas I would assume that the intended behaviour here is to just expose the user-defined variables, and not the raw `env` bindings?
```
{
  "a": "a"
}
```

I can't see any reason to expose the raw DO stub on `env` here? Happy to fix it, just verifying assumptions.

threepointone:
Yeah you’re right, we need to remove the durable object bindings. Please and thank you!

PR #163, opened 2023-07-11
Replace room.getWebSockets with a shimmed room.connections when using Hibernation API

jevakallio:
_Note: This PR is stacked on top of #157_

Replaces the new `room.getWebSockets` with a `room.connections` property that behaves like the `connections` map we currently use in `onConnect`. Along with moving the Room ID back onto the `room` property in https://github.com/partykit/partykit/pull/157/commits/1f0c4c66e9856bf0c503ecd00b1af06a3ec5e22a, this now makes the Connection and Room APIs identical between `onMessage` and `onConnect`.

Opening PR for early comments. This seems to work correctly in dev mode, but I still need to implement the same behaviour on the platform side.

### DIFF
--- a/packages/partykit/facade/source.ts
+++ b/packages/partykit/facade/source.ts
@@ -120,28 +120,16 @@ function createDurable(Worker: PartyKitServer) {
         broadcast: this.broadcast,
       };
 
-      // when using the Hibernation API, we cannot track active connections manually, so we'll lazily
-      // construct the connections map by deserializing the sockets tracked by the platform
       if ("onMessage" in Worker) {
-        let connections: Map<string, PartyKitConnection> | null = null;
-        this.room = {
-          ...this.room,
-          get connections() {
-            // initialize connections map the first time the getter is called
-            // TODO: figure out if this is safe (i.e. will not be cached between connects/disconnects),
-            // or if we need to recreate this manually before every onMessage invocation
-            if (connections === null) {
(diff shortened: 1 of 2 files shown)


packages/partykit/facade/source.ts
```diff
@@ -102,20 +109,40 @@ function createDurable(Worker: PartyKitServer) {
...
+
+      // when using the Hibernation API, we cannot track active connections manually, so we'll lazily
+      // construct the connections map by deserializing the sockets tracked by the platform
+      if ("onMessage" in Worker) {
+        let connections: Map<string, PartyKitConnection> | null = null;
+        this.room = {
+          ...this.room,
+          get connections() {
+            // initialize connections map the first time the getter is called
+            // TODO: figure out if this is safe (i.e. will not be cached between connects/disconnects),
```
jevakallio, line 131: The downside here is that we need to deserialize the socket state for each socket to grab their ids to be used as the Map keys. This could possibly be performance-sensitive.

To address that, I've applied some caching here to make sure we don't have to re-initialize the for each access to the `connections` getter. As per comment, I'm not sure whether this is safe (in case the class instance is shared between `onMessage` invocations, and connections are created/dropped between invocations). 

This seems to work in dev mode, but will need to verify the same approach will work on the platform side.

Whether we can cache this or not, would be good to do some CPU load testing here with very large number of connections.

PR #164, opened 2023-07-12
feat: add support for read-only connections

iojcde:
This PR implements a read-only mode, where a user would not be allowed to make changes to a given room.


threepointone:
Thank you very much, sorry for the delay!

PR #166, opened 2023-07-12
feat: add `ctx` to `onConnect`

threepointone:
## `onConnect(connection, room, context){}`

Developers want access to additional metadata available on every connection. For example, to implement a rate limiter, they need access to the IP address of the client. They may want access to the user agent, some cookie values, or any of the special `cf` properties like geo information, etc. Most of this information is available on the request object when the connection is made, but we don't actually expose that to the developer right now. I propose we do that.

We can expose a third parameter to the `onConnect` method that is the `context` object. For now it'll just be an onject with a single property `request` that is the request object. In the future we can add more properties to this object.

```js
onConnect(connection, room, context) {
  const { request } = context;
  const { headers } = request;
  const ip = headers['x-forwarded-for'];
  // do something with the ip
}
```

Now, the hibernation api variant doesn't include `onConnect` at all. So, we'll need to add that to the hibernation api. So, I also propose that we let folks add an optional `onConnect` method when defining a party. Calling `connection.addEventListener` with that variant will silently do nothing, but maybe we could override that to throw an error instead. Importantly, folks should be able to call `.serializeAttachment` in the `onConnect` method, and recover that data in the `onMessage` method.

```js
export default {
  onConnect(connection, room, context) {
    const { request } = context;
    const { headers } = request;
    const ip = headers["x-forwarded-for"];
    connection.serializeAttachment({ ip });
    // do something with the ip
  },
  onMessage(message, connection, room) {
    const { ip } = connection.deserializeAttachment();
    // do something with the ip
  },
};
```

This PR adds the context object to `onConnect`. After we land the hibernation api, we can do the additional work mentioned.

### DIFF
--- a/.changeset/tough-pandas-shout.md
+++ b/.changeset/tough-pandas-shout.md
@@ -30,7 +30,7 @@ export default {
     connection.serializeAttachment({ ip });
     // do something with the ip
   },
-  onMessage(message, connection, room) {
+  onMessage(connection, message, room) {
     const { ip } = connection.deserializeAttachment();
     // do something with the ip
   },
--- a/.changeset/tough-pandas-shout.md
+++ b/.changeset/tough-pandas-shout.md
@@ -1,0 +1,40 @@
+---
+"partykit": patch
+---
+
+feat: add `ctx` to `onConnect`
+
+## `onConnect(connection, room, context){}`
+
+Developers want access to additional metadata available on every connection. For example, to implement a rate limiter, they need access to the IP address of the client. They may want access to the user agent, some cookie values, or any of the special `cf` properties like geo information, etc. Most of this information is available on the request object when the connection is made, but we don't actually expose that to the developer right now. I propose we do that.
+
+We can expose a third parameter to the `onConnect` method that is the `context` object. For now it'll just be an onject with a single property `request` that is the request object. In the future we can add more properties to this object.
+
+```js
+onConnect(connection, room, context) {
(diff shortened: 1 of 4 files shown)


.changeset/tough-pandas-shout.md
```diff
@@ -0,0 +1,40 @@
...
+```js
+export default {
+  onConnect(connection, room, context) {
+    const { request } = context;
+    const { headers } = request;
+    const ip = headers["x-forwarded-for"];
+    connection.serializeAttachment({ ip });
+    // do something with the ip
+  },
+  onMessage(message, connection, room) {
```
jevakallio, line 33: I believe the signature for `onMessage` is:

```suggestion
  onMessage(connection, message, room) {
```

jevakallio (approved):
Looking good, with one minor fix in docs the changelog.

I'm not 100% convinced about the Hibernation API changes, but since those are not in scope for this PR, we can think about them a little bit longer.  In the long term, it would be nice if the `context` object would be something that we could share between `onConnect`/`onMessage`, and make it survive hibernation by default.  The blocker here is the 2KB limit on the socket attachment, which we don't want to accidentally blow.

.changeset/tough-pandas-shout.md
threepointone: thanks for the catch!

threepointone:
Agreed on all counts. Not so sure about the hibernation version myself.

Issue #161, opened 2023-07-11
Persist `room.storage` locally when in development mode


paulxuca:
Currently `room.storage` seems to disappear when in development mode when re-running the dev command (reload on file change persists correctly). Would be amazing if storage was persisted in development mode when stopping the dev command.

threepointone:
Thanks for filing this! I had a quick look and this might be fairly straightforward, we'll get to it soon.

PR #168, opened 2023-07-15
feat: persist state locally in dev

threepointone:
Currently, `.storage` is only held in memory. This is fine for most usecases, but it means if you shutdown the dev process, we lose any state you may have been holding. This PR adds a config/flag `persist` to store this locally. We also turn it on by default. You can pass `--persist false` in the cli, or `persist: false` in config to turn it off (or `true` for default path). You can also pass `--persist some/custom/path` in the cli, or `persist: "some/custom/path" in the config, to use a custom path for the state.

Fixes https://github.com/partykit/partykit/issues/161

### DIFF
--- a/.changeset/giant-clouds-itch.md
+++ b/.changeset/giant-clouds-itch.md
@@ -1,0 +1,9 @@
+---
+"partykit": patch
+---
+
+feat: persist state locally in dev
+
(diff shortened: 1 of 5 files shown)


threepointone:
landing this, need it locally ha.

PR #169, opened 2023-07-15
run checks on all OSes for PRs

threepointone:
let's see what CI says


threepointone:
tests still failing in windows, but it's already a lot better, so I'm going to comment that bit out and land this PR

PR #171, opened 2023-07-16
Add `once` and `signal` options to PartySocket.addEventListener

tom-sherman:


packages/partysocket/src/ws.ts
```diff
@@ -357,6 +371,12 @@ export default class ReconnectingWebSocket {
     if (listeners) {
       for (const listener of listeners) {
         this._callEventListener(event, listener);
+        if (this._onceListeners.has(listener)) {
```
threepointone, line 374: hmm if the listener has _also_ been added without `{once:true}`, then this would remove that as well

tom-sherman: Tbh I didn't know you could add a listener multiple times!

tom-sherman: I mean I can see that from the implementation here, but I mean with native EventTarget behaviour

tom-sherman: I pushed a fix, but not sure this is the "correct" behaviour when comparing to native listeners. This seems to suggest that you can only add a handler once and subsequent subscriptions are ignored?

image

tom-sherman: But this means the current `addEventListener` implementation of using an array instead of a Set is incorrect 😅

packages/partysocket/src/ws.ts
```diff
@@ -357,6 +373,12 @@ export default class ReconnectingWebSocket {
     if (listeners) {
       for (const listener of listeners) {
         this._callEventListener(event, listener);
+        if (this._onceListeners.get(listener) === 1) {
```
threepointone, line 376: How does it work in the following situation:

```js 
function fn(){ /* ... */}
socket.addListener(fn)
socket.addListener(fn, {once:true))
socket.addListener(fn)
socket.addListener(fn, {once:true))
socket.addListener(fn, {once:true))

```

Do you suppose this deserves some tests?

(alternately, we should just extend EventTarget and get all this for free, but that's a much bigger refactor I don't think we should do now)

packages/partysocket/src/ws.ts
threepointone: yeah the array seems wrong anyway

packages/partysocket/src/ws.ts
tom-sherman: Yeah this should definitely have tests, will add them before moving out of draft.

packages/partysocket/src/ws.ts
threepointone: hrmm we should match the native behaviour

tom-sherman: From your other thread

> alternately, we should just extend EventTarget and get all this for free, but that's a much bigger refactor I don't think we should do now

Probably makes sense to do this instead? Would get guaranteed correct behaviour then.

threepointone:
Afk for a couple of hours, I’ll come back and help out if you need it

tom-sherman:
Closing this in favour of #172.

PR #178, opened 2023-07-17
remove `unstable_initial`, and `onBeforeConnect` can now return a `Request`/`Response`

threepointone:
We were using `unstable_initial` to pass info between `onBeforeConnect` and `onConnect`. But it just felt so alien. Now that onConnect also receives request (via https://github.com/partykit/partykit/pull/166), let's just let onBeforeConnect return a Request or a Response. If a Request, it gets passed on to onConnect. Else, it just returns.

This is a breaking change, but it's a minor one, and it's worth it to get rid of `unstable_initial`.

### DIFF
--- a/packages/partykit/facade/source.ts
+++ b/packages/partykit/facade/source.ts
@@ -198,17 +198,12 @@ function createDurable(Worker: PartyKitServer) {
           }
           connectionId = crypto.randomUUID();
         }
-        const rawInitial = request.headers.get("x-pk-initial");
-        const unstable_initial = rawInitial
-          ? JSON.parse(rawInitial)
-          : undefined;
 
         // TODO: Object.freeze / mark as readonly!
         const connection: PartyKitConnection = Object.assign(serverWebSocket, {
           id: connectionId,
           roomId: this.room.id,
           socket: serverWebSocket,
-          unstable_initial,
         });
         this.room.connections.set(connectionId, connection);
 
@@ -219,7 +214,6 @@ function createDurable(Worker: PartyKitServer) {
           this.controller.acceptWebSocket(serverWebSocket);
           connection.serializeAttachment({
             id: connectionId,
-            unstable_initial,
             roomId: this.room.id,
           });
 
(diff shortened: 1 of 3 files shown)


packages/partykit/facade/source.ts
```diff
@@ -403,14 +398,11 @@ export default {
...
-                headers: {
-                  ...Object.fromEntries(request.headers.entries()),
-                  "x-pk-initial": JSON.stringify(onBeforeConnectResponse),
-                },
-              })
-            );
+            if (onBeforeConnectResponse instanceof Response) {
+              return onBeforeConnectResponse;
+            } else if (onBeforeConnectResponse instanceof Request) {
+              return await env.MAIN_DO.get(id).fetch(onBeforeConnectResponse);
```
jevakallio, line 404: Here we probably need to validate that the Request URL is still pointing to the correct room, e.g. `/party/123`.

At the very least, we'll want to validate that it's pointing to **some** room: I could possibly see a use case where the `onBeforeConnect` handler is used as a router to reroute the request to a different room, e.g. rewriting `/party/123` to `party/abc`, or even a sub-party `parties/other-party/123`. 

In fact, this pattern could be useful for things like load-balancing between multiple rooms, or even migrations!

jevakallio:
This is looking good to me!

We may need to do a little work with the type definitions here. Because `url`, and the request body are read-only properties, the only ways to transmit information from `onBefore*` to `on*` handlers are either to:
- Pass information in request headers by modifying them in-place using `request.headers.set`, or
- Construct a new `Request` object that changes the Request URL or body

To me it would make sense to also support the latter, but AFAICS the typings in `@cloudflare/workers-types` doesn't make this easy. We want to copy all Request fields, including `request.cf` to the new request, but

- `new Request(request)` constructor will be assumed to be a Fetch API `Request`, which doesn't accept a CloudFlare request as the input parameter
- `import { Request } from "@cloudflare/workers-types";` will fail with `"No matching export in "node_modules/@cloudflare/workers-types/index.ts" for import "Request"` because the package exports only types, not the actual constructor
- `import type { Request } from "@cloudflare/workers-types";` will cause a type error `'Request' cannot be used as a value because it was imported using 'import type'.ts`

Currently the only way I can see doing this is `new Request(request as unknown as Request)` which is not only incorrect, but also gross.

This is not a blocker for deprecating `unstable_initial` since that's already using headers as the transmission mechanism anyway, but would be good to solve to expand the `onBefore*` handler capabilities here.

threepointone:
Thanks for the review! Made the changes, landing now.

PR #180, opened 2023-07-18
Install @cloudflare/workers-types transitively

jevakallio:
## Problem

The `PartyKitConnection` type resolves to `any` unless the consumer has installed `@cloudflare/workers-types` definitions:

image

This is because the type extends the `WebSocket` type, which won't be available unless the type definitions are installed:

```
import type { DurableObjectStorage, ExecutionContext, Request, Response, WebSocket } from "@cloudflare/workers-types";

export type PartyKitConnection = WebSocket & {
    id: string;
    /**
     * @deprecated
     */
    socket: WebSocket;
    unstable_initial: unknown;
};
```

## Solution

Move `@cloudflare/workers-types` from `devDependencies` to `dependencies` to ensure they're installed in client projects.

### DIFF
--- a/packages/partykit/package.json
+++ b/packages/partykit/package.json
@@ -12,6 +12,7 @@
     "./src/cli": "./src/cli.ts"
   },
   "dependencies": {
+    "@cloudflare/workers-types": "^4.20230628.0",
     "chalk": "^5.2.0",
     "chokidar": "^3.5.3",
     "clipboardy": "^3.0.0",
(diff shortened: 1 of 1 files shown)


threepointone (approved):
god's work

packages/partykit/package.json
```diff
@@ -12,6 +12,7 @@
     "./src/cli": "./src/cli.ts"
   },
   "dependencies": {
+    "@cloudflare/workers-types": "^4.20230628.0",
```
jevakallio, line 15: I wonder if we should also loosen the version constraint here to make sure we're compatible with other type definitions our users may be using?

I opted to keep this pinned to minor for now since that's these are the correct types for the current runtime.

PR #181, opened 2023-07-18
feat: add support for headers in callback options

iojcde:
When a user needs to add headers to the callback (ex authentication headers), currently they have to write a handler function that requests the receiving server. 

This PR adds an simpler way to implement something of the sort: it adds a `header` option in the interface UrlCallbackOptions` where a user can specify Headers to be added to the callback request.


threepointone (approved):
looks good!

PR #182, opened 2023-07-19
fix: change type of `conn` to PartykitConenection

iojcde:
Currently the type of `conn` for onConnect in y-partykit is WebSocket, which is incompatible with the type definition of onConnect of `partykit/server`.

This PR replaces the type of the connections in `y-partykit` to the correct type.


threepointone (approved):
Good catch, thank you so much for this!

PR #187, opened 2023-07-20
allow using `onConnect` with `onMessage`

threepointone:
We now allow using `onConnect` with `onMessage`. This lets you get access to the context `{request}`. Warning: You can't use `.addEventListener` when using `onMessage`, but it currently just silently fails. We'll make this an error later.


threepointone:
landing to unblock matt

PR #190, opened 2023-07-20
Clean up hibernatable connection before invoking event handler

jevakallio:
Fixes issue where `connections` keeps growing on every connection when opting into Durable Object Hibernation.


threepointone (approved):
🚀

PR #192, opened 2023-07-20
y-partykit: better default for `gc`

threepointone:
It's better to default to `gc: true` with partykit, so memory is better used in partykit servers. We should also ensure `gc` and `persist` are never used together. So this PR adds a little logic to find a better default, and ensuring the 2 never clash.


threepointone:
Tested this, feeling good about it, landing.

PR #201, opened 2023-07-22
fix: correct room id/parties in hibernation mode

threepointone:
We weren't hydrating the room id and `parties` correctly in hibernation mode, this patch fixes that.


packages/partykit/facade/source.ts
```diff
@@ -131,7 +135,8 @@ function createDurable(Worker: PartyKitServer) {
       });
 
       this.room = {
-        id: "UNDEFINED", // using a string here because we're guaranteed to have set it before we use it
+        // @ts-expect-error - id is a symbol when we start
+        id: UNDEFINED, // using a string here because we're guaranteed to have set it before we use it
```
jevakallio, line 139: We were using a string here to avoid making the type optional, but since we're ignoring the type error here anyway, is there a benefit of using a symbol as sigil instead of just plain old `undefined`?

In both cases, because the actual runtime value of the object does not match the types, this is an opportunity to cause a possible runtime bug. In an ideal world, we'd have a different type that we use internally in this file that marks the `id` field as optional, and then assert it to be defined before passing it to the userland code, where the field is non-optional.

jevakallio (approved):
Looks good, minor question/nit on typings.

PR #204, opened 2023-07-26
add a user agent to all our fetch calls

threepointone:
Good manners, and will be useful when we start debugging/o11ying


threepointone:
landing this one

PR #208, opened 2023-07-27
Connect to subparty

jevakallio:
Allows `partysocket` to connect to a party other than the main party, e.g. 
```js
  const socket = usePartySocket({
    host: "localhost:1999",
    party: "chatroom",
    room: roomId,
  });
```


threepointone (approved):
Awesome, ship it!

PR #223, opened 2023-08-01
DO NOT LAND feat: use `PartySocket` when connecting via `room.parties`

threepointone:
This uses `PartySocket` when connecting why `room.parties`, whcih gives you reconnection/buffering logic without any extra effort. Non-breaking change, existing code should still just work.


threepointone:
closing for https://github.com/partykit/partykit/pull/395

PR #224, opened 2023-08-01
enable inspector in non-tests 😖

threepointone:
In https://github.com/partykit/partykit/pull/222, we didn't actually enble the inspector outside of tests. Oops. This fixes that.


sylwiavargas (approved):
thank you!

PR #228, opened 2023-08-02
fix: properly sort chunking in y-partykit for big values

threepointone:
So, out persistence layer has a 128kb limit on values, but `y-partykit` should/could be used for documents bigger than that. So we implemented a chunking strategy that break up values across multiple keys. When making keys for these, we didn't pad the generated indexes, which meant that for sufficiently large values, we might have assembled them back in the wrong order (because lexicographical sorting). This is a fix for that. It's a breaking change, but oddly I haven't heard from people who've faced the problem at all.

### DIFF
--- a/.changeset/fuzzy-geese-yawn.md
+++ b/.changeset/fuzzy-geese-yawn.md
@@ -1,0 +1,7 @@
+---
+"y-partykit": patch
+---
+
+fix: properly sort chunking in y-partykit for big values
+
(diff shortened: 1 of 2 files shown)


jevakallio (approved):
LGTM! Nice catch!

threepointone:
This might break existing projects, so I'm a bit concerned. Wondering if there's a workaround / fix / migration...

threepointone:
Ok I tested it and it didn't seem to break anything? Landing this now

PR #233, opened 2023-08-06
check if crypto is defined

turkerdev:
if we don't check if crypto is defined or not.
it gives the following error.

```
ReferenceError: crypto is not defined
```


threepointone:
Great catch, thank you very much!

PR #239, opened 2023-08-12
partysocket: warn if EventTarget is not available

threepointone:
Log an error if `EventTarget` isn't available when using `partysocket`, with instructions on fixing it.


threepointone:
cc @brentvatne @tom-sherman

PR #258, opened 2023-08-16
Migrate to Clerk authentication behind a feature flag

jevakallio:
Branching this out of #221 to make it easier to review.

- Implements a very simple feature flagging system.
  - Caches previously used flags on disk
  - Refreshes flags lazily on every use to avoid blocking, so new flags will only be applied on next use
  - Falls back to default flags
- Chooses login mechanism based on feature flag
  - Defaults to `github` for now, we'll change it to `clerk` on launch day
  - On launch day + `n`, we can remove `github` from supported platforms, which will force all existing clients to migrate
- Adds a `--provider` argument to `partykit login` for early testing (to be removed once stable)

### DIFF
--- a/packages/partykit/src/bin.tsx
+++ b/packages/partykit/src/bin.tsx
@@ -293,11 +293,17 @@ envCommand
 program
   .command("login")
   .description("Login to PartyKit")
-  .action(async () => {
+  .addOption(
+    new Option(
+      "-p, --provider <provider>",
+      "login provider (experimental)"
+    ).choices(["github", "partykit"])
+  )
+  .action(async ({ provider }: { provider?: "github" | "partykit" }) => {
     await printBanner();
     render(
       <Suspense>
-        <Login />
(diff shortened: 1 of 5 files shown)


packages/partykit/src/bin.tsx
```diff
@@ -293,11 +293,17 @@ envCommand
...
   .command("login")
   .description("Login to PartyKit")
-  .action(async () => {
+  .addOption(
+    new Option(
+      "-p, --provider <provider>",
+      "login provider (experimental)"
+    ).choices(["github", "partykit"])
+  )
+  .action(async ({ provider }: { provider?: "github" | "partykit" }) => {
```
jevakallio, line 302: Calling this `partykit` just so we don't have to put the "clerk" name out to the docs. This will be removed within a few weeks in either case.

threepointone (approved):
perfect. thanks for making flag fetching non-blocking!

PR #261, opened 2023-08-17
Class API for PartyKit server

jevakallio:
This PR implements a barebones version of a class-based API for PartyKit.

## In scope

- [x] Support Class API
- [x] Implement `onStart`
- [x] Implement `options`
- [x] Move `room.parties` to `party.context.parties`
- [x] Call `onMessage`, `onClose`, `onError` for non-hibernating connections 
- [x] Fix `room.id` and `room.parties` being undefined in contructor/`onStart` 
- [x] Make `connections` lazy 
  - [x] Tag connections
- [x] Rename/alias type exports, deprecate old types
- [x] Make `onFetch` stable
- [x] Documentation in changeset
- [x] Update PartyWorker implementation

## Follow-on in future PRs
- [ ] Write tests
- [ ] Better runtime validation for handlers


examples/class/src/server.ts
```diff
@@ -0,0 +1,45 @@
...
+  Party,
+  PartyKitConnection,
+  PartyKitRequest,
+  PartyKitResponse,
+  PartyServer,
+  PartyServerConstructor,
+  PartyServerOptions,
+} from "partykit/server";
+
+export default class Main implements PartyServer {
```
threepointone, line 11: how are  you feeling about `implements` vs `extends`?

packages/partykit/src/server.ts
```diff
@@ -132,6 +133,79 @@ type ConnectionHandler = RequestHandler & {
 
 export type PartyKitServer = ConnectionHandler;
 
+// New Class API
```
threepointone, line 136: the word PartyKit comes up everywhere haha. How do you feel about going full ios and start using PK? Or do you feel comomfotable with the full word?

threepointone (commented):
I can't wait, this has been sorely need for months now. I'm going to start playing with this over the weekend.

packages/partykit/src/server.ts
jevakallio: I'm thinking `Party*` (drop the "Kit", it's cleaner) rather than `PK` (never go full NextStep)

Part of me just wants to go prefixless, e.g. just `Connection`, `Server`, etc. Where this falls apart is `Request` and `Response`, which cannot be imported as types since they squat the global Request/Response constructors.

So the two options I'd consider would be:
```
import type {
  Party,
  PartyRequest,
  PartyResponse,
  Connection,
  Server,
  ServerOptions,
} from "partykit/server";
```

Or slightly longer, but with more naming parity:
```
import type {
  Party,
  PartyRequest,
  PartyResponse,
  PartyConnection,
  PartyServer,
  PartyServerOptions,
} from "partykit/server";
```

Honestly it sucks that we need all these imports, but I don't know that there is any way around this.

I'll continue to investigate if we can drop the `Request` / `Response` imports in the default case by having a global declaration for those types, in which case we could do:

```
import type {
  Party,
  Connection,
  Server,
  ServerOptions,
} from "partykit/server";
```

Which is probably the best we can do.

examples/class/src/server.ts
jevakallio: Leaning towards `implements`, though I'm not 100% decided on it yet.

packages/partykit/facade/source.ts
```diff
@@ -176,24 +221,22 @@ function createDurable(Worker: PartyKitServer) {
...
-        });
-
-        // populate the room id/slug if not previously done so
 
-        const roomId = getRoomIdFromPathname(url.pathname);
-
-        assert(roomId, "No room id found in request url");
-        this.room.id = roomId;
+      try {
+        if (!this.worker) {
```
jevakallio, line 231: There was a bit of a conceptual issue with the new `constructor` / `onStart` "lifecycle" methods: they happen before the first request arrives, so we don't know the room id yet.

```
export default class Main implements PartyServer {
  constructor(readonly party: Party) {
    console.log("Constructor", party.id);
  }

  async onStart() {
    console.log("onStart", this.party.id);
  }

  async onConnect(connection: PartyKitConnection) {
    console.log("onConnect", this.party.id);
  }
}
```

Outputs:
```
Constructor undefined
onStart undefined
onConnect some-room
```

The obvious solution here was to move the `PartyServer` class initialization into `fetch`/`webSocketMessage`.

Might be easier to review this change by reading the whole commit:
https://github.com/partykit/partykit/pull/261/commits/73a42d57426c6c1ddd609759aed179c51b86f212

packages/partykit/facade/connection.ts
```diff
@@ -0,0 +1,189 @@
...
+    get uri() {
+      return deserializeAttachment(ws).uri;
+    },
+    get socket() {
+      return ws;
+    },
+  });
+};
+
+class HibernatingConnectionIterator
```
threepointone, line 43: yo this slaps

packages/partykit/facade/source.ts
threepointone: reviewed, looks good. smart.

threepointone (approved):
This is fantastic, was looking forward to this. Lemme know when you're sharing a fleshed out changeset or doc!

PR #285, opened 2023-08-24
Make YProvider work in workers runtime

jevakallio:
- Fix unnecessary `Y.js` import
- Polyfill missing `WebSocket.[STATUS]` codes
  - This change could also have been made by checking for `WebSocketStatus` explicitly, but because the JS standard WebSocket type definition claims they exist, I'm 100% sure this will bite other people as well
-  Disable Broadcast Channel by default

### DIFF
--- a/packages/y-partykit/src/provider.ts
+++ b/packages/y-partykit/src/provider.ts
@@ -25,7 +25,6 @@ export const WebSocketStatus = {
 // Polyfill WebSocket status code constants for environments that don't have them
 // (e.g. Cloudflare Workers)
 if (WebSocket.OPEN === undefined) {
-  console.log("Polyfilling WebSocketStatus");
   Object.assign(WebSocket, WebSocketStatus);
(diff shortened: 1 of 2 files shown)


packages/y-partykit/src/provider.ts
```diff
@@ -15,6 +15,23 @@ export const messageQueryAwareness = 3;
...
+export const WebSocketStatus = {
+  CONNECTING: 0,
+  OPEN: 1,
+  CLOSING: 2,
+  CLOSED: 3,
+};
+
+// Polyfill WebSocket status code constants for environments that don't have them
+// (e.g. Cloudflare Workers)
+if (WebSocket.OPEN === undefined) {
```
threepointone, line 27: Should we do this in our source.ts facade? Maybe also file an issue for workerd?

jevakallio: Yes, and yes, and also in the PartyKit platform facade. Will do all three.

PR #296, opened 2023-08-29
Generate session tokens for CI

jevakallio:
Adds a new `npx partykit token generate` command that allows you to generate a new Clerk client session.

From code perspective, this is ready to land, but since it relies on the yet-unpublicised dashboard, it may make sense to only ship this when we're ready to onboard users there.

While developing this, I made a few other improvements to the session management worflow:
- Sessions are now tagged with custom user agents, so the user can differentiate between CLI sessions and (CI) token sessions in the Clerk account management page
- Sessions are expired upon logout (because our device session length is very long, these will otherwise pile up)

### DIFF
--- a/packages/partykit/src/auth/device.ts
+++ b/packages/partykit/src/auth/device.ts
@@ -12,7 +12,7 @@ if (!DASHBOARD_BASE) {
   throw new Error("PARTYKIT_DASHBOARD_BASE not defined");
 }
 
-export async function signInWithBrowser(): Promise<
+export async function signInWithBrowser(mode: "cli" | "token"): Promise<
   | {
       token: string;
       teamId: string;
@@ -88,7 +88,7 @@ export async function signInWithBrowser(): Promise<
     const redirectUrl = `http://localhost:${port}/device/callback`;
 
     // the remote server url that will handle the login flow
-    const loginUrl = `${DASHBOARD_BASE}/login/device?redirectUrl=${encodeURIComponent(
+    const loginUrl = `${DASHBOARD_BASE}/login/device?mode=${mode}&redirectUrl=${encodeURIComponent(
       redirectUrl
     )}`;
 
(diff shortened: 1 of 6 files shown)


packages/partykit/src/auth/device.ts
```diff
@@ -88,7 +88,7 @@ export async function signInWithBrowser(): Promise<
     const redirectUrl = `http://localhost:${port}/device/callback`;
 
     // the remote server url that will handle the login flow
-    const loginUrl = `${DASHBOARD_BASE}/login/device?redirectUrl=${encodeURIComponent(
+    const loginUrl = `${DASHBOARD_BASE}/login/device?mode=${mode}&redirectUrl=${encodeURIComponent(
```
jevakallio, line 91: Sending the mode so we can vary the UI on the dashboard side. Not making those changes now, just preparing for them.

packages/partykit/src/config.ts
```diff
@@ -169,9 +185,36 @@ export async function fetchUserConfig(
...
 export async function logout() {
-  if (fs.existsSync(USER_CONFIG_PATH)) {
+  const config = readUserConfig(USER_CONFIG_PATH);
+  if (config) {
     fs.rmSync(USER_CONFIG_PATH);
+    if (config.type === "clerk") {
+      try {
+        await expireClerkClientToken(config.access_token);
+      } catch (e) {
+        logger.info(chalk.dim("You were logged out this device."));
```
jevakallio, line 215: This is stupid... it really doesn't matter that the session is not deleted, so we don't want to show a big scary warning.

But at the same time I don't want to silently ignore this error, in case something breaks in the implementation, we should notice it.

What we really need to do is integrate an error reporting service to push these silent errors into.

threepointone (approved):
hell yeah lgtm

PR #304, opened 2023-08-31
Convert `create-partykit` template to Class API

jevakallio:
- [x] Convert  `create-partykit` template to Class API
- [x] Convert `init`
- [x] Use `client.(ts|js)` instead of `jsx|tsx`
- [x] Fix JSDoc types
- [x] Remove unnecessary assets


packages/create-partykit/js-template/public/index.html
```diff
@@ -8,12 +8,8 @@
     />
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <title>PartyKit: Everything's better with friends!</title>
-    <!-- Icons / Favicon -->
+    <!-- Favicon -->
     <link rel="icon" href="/favicon.ico" sizes="any" />
-    <link rel="icon" href="/logo.png" type="image/png" />
```
jevakallio, line 13: This commit might be a little controversial. I don't think the template project needs all the platform-specific favicons. It makes the initial deploy slower and feel clunkier, and if anyone really wants to build a production-capable application, they can add these themselves.

@threepointone let me know if you disagree, I can revert the last commit to add these back, but this would be my preference.

packages/create-partykit/js-template/src/server.js
```diff
@@ -1,43 +1,47 @@
 /* eslint-env browser */
-/** @typedef {import("partykit/server")} PK */
+
+// @ts-check
```
jevakallio, line 3: We didn't actually run `@ts-check` before. Once I turned it on turns out the type imports weren't working, so nothing was being checked.

Personally, I wouldn't bother with JSDoc types for the JavaScript templates at all. What are the chances that someone wants plain js WITH type checking, since the TypeScript bundling is built in.

I'd cater the JS template for "dynamic type enthusiasts" aka daredevils and remove these type checks altogether, but I didn't want to make that change here.

packages/create-partykit/js-template/public/index.html
threepointone: I don't disagree at all. I just picked it from a boilerplate myself. Thanks for removing them.

packages/create-partykit/js-template/src/server.js
threepointone: I'm actually not sure what to do here. I keep hearing conflicting opinions on js only, js with types, ts only. I did the jsdoc version but my heart wasn't really in it. I'm happy for you to take a call here. I assume most people will just use ts anyway.

threepointone (approved):

packages/create-partykit/js-template/src/server.js
jevakallio: Thanks, landing this with the jsdoc types since they work now -- they're easier to remove than to add back, let's reconsider if we hear any feedback to the contrary.

PR #305, opened 2023-08-31
Fix edge case bugs in init

jevakallio:
Few bugs I found while fixing #304

### DIFF
--- a/packages/partykit/src/cli.tsx
+++ b/packages/partykit/src/cli.tsx
@@ -168,7 +168,7 @@ export async function init(options: {
         // so let's add the partykit dependency
         // and make a partykit.json file
 
-        if (!packageJson.devDependencies.partykit) {
+        if (!packageJson.devDependencies?.partykit) {
           packageJson.devDependencies = packageJson.devDependencies || {};
           packageJson.devDependencies.partykit = packageVersion;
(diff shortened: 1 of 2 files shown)


packages/partykit/src/cli.tsx
```diff
@@ -168,7 +168,7 @@ export async function init(options: {
         // so let's add the partykit dependency
         // and make a partykit.json file
 
-        if (!packageJson.devDependencies.partykit) {
+        if (!packageJson.devDependencies?.partykit) {
```
jevakallio, line 171: Fails if package.json doesn't have dev deps

packages/partykit/src/cli.tsx
```diff
@@ -214,7 +214,7 @@ export async function init(options: {
       }
 
       if (!options.dryRun) {
-        fs.mkdirSync(path.join(process.cwd(), "src"), { recursive: true });
+        fs.mkdirSync(path.join(process.cwd(), "party"), { recursive: true });
```
jevakallio, line 217: This seems a typo / bad refactor, we're not using `src` directory.

threepointone (approved):
I looked at this a lot and tested it throughly, thanks so much

PR #309, opened 2023-09-01
remove `./` from `create-partykit` path placeholder

austincrim:
Tiniest of nits, but I stubbed my toe here when running `npm create partykit` because the placeholder shows a path with `./` but entering a path starting with `./` will throw an error. If you actually do want to accept `./`, feel free to take this as a bug report and close the PR.

CleanShot 2023-08-31 at 20 58 31@2x

CleanShot 2023-08-31 at 20 57 38@2x

### DIFF
--- a/packages/create-partykit/src/index.tsx
+++ b/packages/create-partykit/src/index.tsx
@@ -194,7 +194,7 @@ export async function init(options: {
           <Text>Where should we create your project?</Text>
           <TextInput
             value={text}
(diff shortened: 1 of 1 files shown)


threepointone:
Hmm we should be smarter here. The `./` is an indicator that it's a path, but we should accept input both with and without. This is a good nit, thanks for filing it. We'll do something about it.

PR #312, opened 2023-09-01
Move website into main repo

threepointone:
(No need to do a review, just tagging so you know it's happening)


jevakallio:
Awesome! I'll move the docs and blog in next week when the content is more mature, now it's still beneficial to be able to iterate directly on `main`.

If you're doing DNS changes for the main site, feel free to set up the `blog.` and `.docs.` subdomains to point to the vercel sites as well, I think those will remain stable.

threepointone:
Awesome yeah. Yeah I figured I'd do the site so I understand how to do the DNS stuff and prepare for blog/docs when it happens

PR #318, opened 2023-09-04
fix(partysocket): Export ReconnectingWebSocket

sdnts:
This fixes an issue with the generated types. Without this, for some reason TypeScript thinks that the ReconnectingWebSocket type is the PartySocket type.

I reported this in #319  (along with a repro). With this in place, I can use the `ReconnectingWebSocket` type in place of `WebSocket` to fix things, like so: https://codesandbox.io/p/sandbox/purple-cache-cggdd4?file=/src/App.tsx:11,3 (I've verified that this does work)

I would be interested to learn why this is needed though if anyone knows.


threepointone:
That's... odd. I'll have a look at this. Thank you very much for filing both an issue and a possible fix!

packages/partysocket/src/index.ts
```diff
@@ -87,3 +87,4 @@ export default class PartySocket extends ReconnectingWebSocket {
 }
 
 export const WebSocket = ReconnectingWebSocket;
+export { ReconnectingWebSocket };
```
threepointone, line 90: ok I tested, and this works, and doesn't require the new export. 

```suggestion
export { ReconnectingWebSocket as WebSocket };
```

.changeset/rotten-peaches-teach.md
```diff
@@ -0,0 +1,5 @@
+---
+"partysocket": patch
+---
+
+Export ReconnectingWebSocket
```
threepointone, line 5:
```suggestion
Export `WebSocket` from `partysocket` correctly
```

threepointone (approved):

PR #329, opened 2023-09-07
change naming of /server exports

threepointone:
This changes the naming of our exported types from `partykit/server`. The motivation here was aesthetic/simplification. Most of the exports los the `Party` prefix; i.e. `PartyServer` becomes `Server`, `PartyConnection` becomes `Connection`, and so on. If you look at the class example, this doesn't drastically change the code; the import becomes a lot shorter (`import type * as Party from 'partykit/server'` gets all the required types) and required types are taken by `Party.*`.

I also did a run on all the docs/templates and the blog post, but lemme know if I missed anything!

### DIFF
--- a/.changeset/lovely-tigers-marry.md
+++ b/.changeset/lovely-tigers-marry.md
@@ -1,0 +1,11 @@
+---
+"partykit": patch
+"@partykit/blog": patch
+"@partykit/docs": patch
+---
+
+change naming of /server exports
+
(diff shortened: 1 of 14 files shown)


threepointone:
I think instead of aliasing, I should duplicate the type definitions, so cmd click on a type doesn't confuse

jevakallio (approved):
Looking nice and pithy -- I love it!

I pushed a commit to use aliases for the deprecations instead of duplicating the implementation. Because we are not forking the implementation side, if we make API changes to the types, the deprecated types should also receive them
https://github.com/partykit/partykit/pull/329/commits/e2911fb827955ea387b3eff9ee573b025a84f4a0

threepointone:
That’s good, I like it, thanks for pushing to it!

PR #333, opened 2023-09-08
fix: `init` doesn't install the right versions of packages

threepointone:
This fix should pick up the latest versions of partykit /partysocket when installing. We were using the previous beta logic when every package was on the same version. This patch picks up the latest versions of both and uses it.

(also sneaking in a quick types fix for usePartySocket)

### DIFF
--- a/packages/partykit/src/cli.tsx
+++ b/packages/partykit/src/cli.tsx
@@ -120,6 +120,35 @@ export async function init(options: {
 
   const pkgManager = detectPackageManager();
 
+  let latestPartyKitVersion = "*";
+  let latestPartySocketVersion = "*";
+  if (packageVersion.startsWith("0.0.0-")) {
+    // this means it's a beta, so let's use this version everywhere
+    latestPartyKitVersion = packageVersion;
+    latestPartySocketVersion = packageVersion;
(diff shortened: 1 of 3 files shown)


packages/partykit/src/cli.tsx
```diff
@@ -120,6 +120,35 @@ export async function init(options: {
 
   const pkgManager = detectPackageManager();
 
+  let latestPartyKitVersion = "*";
+  let latestPartySocketVersion = "*";
+  if (packageVersion.startsWith("0.0.0-")) {
```
jevakallio, line 125: Is this the best way to detect beta? We may have e.g. `0.1.0-hash` betas down the line?

jevakallio (approved):
Looks great, shipit!

packages/partykit/src/cli.tsx
threepointone: We explicitly prefix with `0.0.0-` fr beta builds here https://github.com/partykit/partykit/blob/8330c5755b88a261cb141993910ce29ac9de7d89/.github/version-script.js#L19, opting out of any form of semver, so it _should_ be fine

PR #345, opened 2023-09-11
add a lint rule for not using deprecated APIs

threepointone:


threepointone:
CI is failing because we keep going beyond the 5 min limit. We could increase the limit, but I think a better solution os to just land #344

PR #339, opened 2023-09-10
partykit: bundle dependencies

threepointone:
This brings back packaging of dependencies for partykit (and introduces it for create-partykit). We had previously disabled it for deps because of esm/cjs nonsense, but we've now figured out how to make them play decently together.

I had to fork ink-testing-library because of https://github.com/vadimdemedes/ink-testing-library/pull/23, I'll remove it once that PR is merged.

This also updates ink to 4.4.1, which fixes our previous issue where it was exiting early.

PR #348, opened 2023-09-11
Fix Clerk login

jevakallio:
This was broken by #339 -- the Clerk import is now correctly understood as a default import, so the previous workaround is no longer necessary.


threepointone (approved):
Sorry for breaking it, thanks for the fix!

PR #349, opened 2023-09-11
don't compile apps in CI

threepointone:
we only compile apps because we need .astro/types to pass CI, so let's just commit them and see what happens


threepointone:
landing this, gets back another 30 seconds

PR #351, opened 2023-09-12
Change serveSinglePageApp to singlePageApp in docs

alexanderson1993:
As I was working on my PartyKit app, I tried setting the `serveSinglePageApp` setting as it was demonstrated in the docs.

As it turns out, this isn't how the config is implemented in the code though. Setting it to `singlePageApp` worked correctly.

This PR changes the docs to match the correct behavior.


threepointone:
Good catch thank you!!

PR #352, opened 2023-09-12
DO NOT LAND (internal) use bun install for deps

threepointone:
blocked on https://github.com/lovell/sharp/issues/3511


threepointone:
closing this. we can't land this without windows support.

PR #361, opened 2023-09-12
Fix `./` paths and other relative paths for create-partykit

### LINKED ISSUE #341: error with cli and ./ file path
Not a biggie, but when prompted for file name, suggested has ./*

So naturally I would do like ./new-app but it errors

```
Invalid path. Please use a path like "randomName" or just "." to use the current directory. You entered "./test-party"
/Users/name_here/.npm/_npx/cadd21533ff9f374/node_modules/create-partykit/src/index.tsx:182
        reject(new Error("Invalid path"));
               ^


Error: Invalid path
```

jevakallio:
Fixes #341 and closes #309. Also expands the legal paths to any relative path from the current working directory.

Valid paths:
```
.
./
../
./hello
../hello
../../hello
hello
hello/world
```

Invalid paths
```
hello world
./hello world
hello.json
/hello

```

As per last invalid path example, we still don't allow absolute paths. We could enable this as well, but it would have required changing the path resolution code elsewhere, and writing to absolute paths is most likely unintended, so I didn't enable it for now.

### DIFF
--- a/packages/create-partykit/src/index.tsx
+++ b/packages/create-partykit/src/index.tsx
@@ -206,7 +206,7 @@ export async function init(options: {
             `Absolute paths not supported. Please use a relative path like "randomName", "./randomName", or just "." to use the current directory. You entered "${text}"`
           )
         );
-        reject(new Error("Invalid path"));
+        reject(new Error("Absolute paths not supported"));
       }
 
       const parsed = path.parse(text);
--- a/packages/create-partykit/src/index.tsx
+++ b/packages/create-partykit/src/index.tsx
@@ -198,14 +198,26 @@ export async function init(options: {
       return;
     }
 
-    function done(text: string) {
-      // TODO: make sure text is a valid path
-      const isValidPathRegex = /^([a-zA-Z0-9_-]+\/)*[a-zA-Z0-9_-]+$/;
-
-      if (!isValidPathRegex.test(text || randomName)) {
+    function done(input: string) {
+      const text = input || randomName;
+      if (path.isAbsolute(text)) {
         console.log(
           chalk.red(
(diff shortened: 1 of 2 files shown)


packages/create-partykit/src/index.tsx
```diff
@@ -198,14 +198,25 @@ export async function init(options: {
...
+        );
+        reject(new Error("Absolute paths not supported"));
+      }
 
-      if (!isValidPathRegex.test(text || randomName)) {
+      const parsed = path.parse(text);
+      const isValidPathRegex = /^\.*?([a-zA-Z0-9_-]{0,}\/)*[a-zA-Z0-9_-]+$/;
+      const isValidPath = (path: string) =>
+        path === "." || path === ".." || isValidPathRegex.test(path);
+      if (!isValidPath(parsed.base) || !isValidPath(parsed.name)) {
```
jevakallio, line 216: Parsing the base path and the base name separately to make the regex simpler

packages/create-partykit/src/index.tsx
```diff
@@ -198,14 +198,25 @@ export async function init(options: {
...
+          )
+        );
+        reject(new Error("Absolute paths not supported"));
+      }
 
-      if (!isValidPathRegex.test(text || randomName)) {
+      const parsed = path.parse(text);
+      const isValidPathRegex = /^\.*?([a-zA-Z0-9_-]{0,}\/)*[a-zA-Z0-9_-]+$/;
+      const isValidPath = (path: string) =>
+        path === "." || path === ".." || isValidPathRegex.test(path);
```
jevakallio, line 215: Checking for relative path parts separately to make the regex simpelr

threepointone (approved):
Looks great, thanks for the fix!

PR #364, opened 2023-09-12
Docs/hibernation

sylwiavargas:
This PR adds a page on Hibernation API


apps/docs/src/content/docs/guides/scaling-partykit-servers-with-hibernation.mdx
```diff
@@ -0,0 +1,163 @@
...
+```
+
+## Benefits
+
+Hibernation allows a single party instance to handle vastly more active connections than it otherwise would:
+
+- Without Hibernation: up to **100 connections** per party instance
+- With Hibernation: up to **32,000 connections** per party instance
+
+The number of working connections may vary depending on the performance characteristics of the code you run on PartyKit (for example, memory use or CPU time). In PartyKit apps, memory use depends on:
```
jevakallio, line 29:
```suggestion
Note that even with Hibernation, the practical maximum amount of connections may vary depending on the performance characteristics of the code you run on PartyKit (for example, memory use or CPU time). In PartyKit apps, memory use depends on:
```

apps/docs/src/content/docs/guides/scaling-partykit-servers-with-hibernation.mdx
```diff
@@ -0,0 +1,163 @@
...
+
+- active connections,
+- code size - for example, third-party libraries,
+- state - class fields you keep in memory between requests,
+- dynamic allocations - memory used by JavaScript for local variables, execution stack, and others.
+
```
jevakallio, line 35: I'd add here something like "With Hibernation, the active WebSocket connections no longer count towards your party's 128MB memory limit, but the total memory footprint of your code, state and dynamic allocations may still exceed that limit."

apps/docs/src/content/docs/guides/scaling-partykit-servers-with-hibernation.mdx
```diff
@@ -0,0 +1,163 @@
...
+Hibernation will work better with certain programming patterns.
+
+### Partial state loading
+
+Instead of loading the full state from storage in the `onStart` lifecycle method, only read from storage the data you need, and only when you need it. Additionally, store state under multiple keys instead of one.
+
+If your party instance requires state, you will need to persist it to party storage, an external database, or an API, and reload it when the party is woken up from hibernation.
+
+:::tip[PartyKit `storage`]
```
jevakallio, line 83: I think the `:::` callouts don't support backtick code, so this renders just as "PartyKit"
```suggestion
:::tip[PartyKit storage]
```

apps/docs/src/content/docs/guides/scaling-partykit-servers-with-hibernation.mdx
```diff
@@ -0,0 +1,163 @@
...
+```ts
+type AllItems = Record<string, any>;
+
+export default class Server implements Party.Server {
+  options: Party.ServerOptions = { hibernate: true };
+  constructor(readonly party: Party.Party) {}
+  // GOOD: no data loading on startup
+  // onStart() {}
+
+async onMessage(websocketMessage: string) {
```
jevakallio, line 102: The code indentation in this code block looks a little off?

apps/docs/src/content/docs/guides/scaling-partykit-servers-with-hibernation.mdx
```diff
@@ -0,0 +1,163 @@
...
+        ...event.data,
+      };
+    };
+
+    // BAD
+    // we need to persist the whole state again (expensive)
+    await this.party.storage.put("items", this.items);
+  };
+};
+````
```
jevakallio, line 160: One too many backticks here (should be 3). Can't do a suggestion, because GitHub uses triple backtick as the suggestion format 😂

jevakallio (approved):
This is very nice! I left a few comments and suggestions.

More broadly, would be good to standardise on some text formatting things. I noticed this page uses:
- Single dashes as thought separators, I prefer to use `—` em dashes
- Single quotes as highlighting (e.g. `'infrequent writes'`), I would prefer using markdown formatting (`**infrequent writes**`)
- Lists items sometimes with lowercase letters and sometimes uppercase. I don't mind either as long as it's consistent, but I would prefer uppercase.

However, I'm not too religious about any of them (except maybe the single quote highlights, that seems a bit weird?), so happy to standardise either way.

apps/docs/src/content/docs/guides/scaling-partykit-servers-with-hibernation.mdx
sylwiavargas: I noticed that Prettier inserts an additional backtick and messes up indentation in .md 😭 Thank you.

apps/docs/src/content/docs/guides/scaling-partykit-servers-with-hibernation.mdx
sylwiavargas: Thank you!

PR #372, opened 2023-09-13
Reduce double ping in client.js/ts

jamesgpearce:
When server hot-reloads (for example), the clients reconnect and another ping sequence starts. You then get multiple pings from each browser. This change cancels an existing sequence if it exists.


threepointone:
Great catch, thank you very much @jamesgpearce!

threepointone (approved):

packages/create-partykit/ts-template/src/client.ts
jamesgpearce, line 1: Weird, number satisfied TS locally.

PR #373, opened 2023-09-14
use npm install instead of npm ci

threepointone:
we have a package-lock, so I'm curious whether it affects install times


threepointone:
no change. bummer.

PR #379, opened 2023-09-16
Compress the header horizontally on small screens

genmon:
There's a bug on mobile where the header can't compress horizontally, and this adds a right-hand margin for the rest of the site. Existing:

image

This PR allows the header to shrink down just enough to avoid this happening:

image

It also removes a rogue right-hand black border on the hero iframe.


threepointone (approved):
Brilliant, thanks for the weekend PR!

PR #387, opened 2023-09-18
Fix static asset deployment

jevakallio:
Seems like the static asset deployment wasn't correctly sending the `preview` parameter.


threepointone:
Good catch, was just looking at this. Thank you!!!

PR #389, opened 2023-09-18
Fix deleting preview deployments

jevakallio:
While fixing #387,  I noticed that deleting preview deployments fails because the `preview` field is not allowed on `Config`.

This makes sense, because the `preview` field is a documented option on partykit json, but it can be provided as a config override with a `--preview` flag. 

I could have fixed this by removing the `preview` flag before we validate the config, but I think it would be useful to be able to set the `preview` flag in partykit.json, so I added it to the valid config schema instead.

### DIFF
--- a/packages/partykit/src/config-schema.ts
+++ b/packages/partykit/src/config-schema.ts
@@ -29,6 +29,7 @@ export const schema = z
     name: z.string().optional(),
     main: z.string().optional(),
     port: z.number().optional(),
+    preview: z.string().optional(),
     serve: z
       .union([
         z.string(),
--- a/.changeset/quiet-mice-chew.md
+++ b/.changeset/quiet-mice-chew.md
@@ -1,0 +1,5 @@
+---
+"partykit": patch
+---
+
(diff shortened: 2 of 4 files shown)


packages/partykit/src/config-schema.ts
```diff
@@ -29,6 +29,7 @@ export const schema = z
     name: z.string().optional(),
     main: z.string().optional(),
     port: z.number().optional(),
+    preview: z.string().optional(),
```
threepointone, line 32: it shouldn't be in config, it can't be used inside partykit.json for example 
but hmm we have it as strict don't we

threepointone: I guess we can remove .strict()
(that or strip preview before passing to getConfig)

threepointone: ok I read your PR description now where you described this ha

threepointone (approved):
let's land it. we shouldn't encourage people to pass preview in partykit.json, so let's just not document it there

PR #407, opened 2023-09-22
Update CI guide

tsriram:
Updated the CI guide based on my experience settings this up. It looks like you can't create a repository secret with name starting with `GITHUB_` (error message below for reference) 

image

Also tweaked it to link to the page where you can create PAT, which is what I did.

Sample workflow which works for me - https://github.com/tsriram/svelte-party-demo/blob/321a8445bbf7322e8ec63962690922729fe1f510/.github/workflows/deploy-partykit.yml

Sample Action run for reference - https://github.com/tsriram/svelte-party-demo/actions/runs/6276609049

PS: Feel free to nitpick on the copy / edit it directly :)


threepointone:
Woah that's weird, we use GITHUB_ for some of our own stuff, this sounds like a new rule they're rolling out. Thank you so much for investigating and this PR!

threepointone (approved):

PR #409, opened 2023-09-23
[fix] filePath in windows

### LINKED ISSUE #328: 404 on Windows 11
Hi,
Got some 404 not found errors on windows for both client.js and client.css. Both files do exist but somehow are not found.
Below is the command line's console log:

```
🎈 PartyKit v0.0.7
-------------------
Build succeeded, starting server...
kj/async-io-win32.c++:982: warning: Bind address resolved to multiple addresses.  Only the first address will be used.  If this is incorrect, specify the address numerically.  This may be fixed in the future.; addrs[0].toString() = 127.0.0.1:49289
[pk:inf] Ready on http://127.0.0.1:1999 
[pk:inf] GET / 200 OK (11ms)
[pk:inf] GET /dist/client.css 404 Not Found (1ms)
[pk:inf] GET /dist/client.js 404 Not Found (2ms)
[pk:inf] GET /normalize.css 200 OK (9ms)
[pk:inf] GET /favicon.ico 200 OK (2ms)
```
Note that Firefox shows the error message below:
`Loading module from “http://127.0.0.1:1999/dist/client.js” was blocked because of a disallowed MIME type (“text/plain”).`
Maybe somehow the server on windows is not using the proper mime types? 
I only have a Windows machine so I do not know whether this happens on Mac/Linux ...etc!

Any idea?

I had a look in RegEdit (Remember I am on Windows) and did not find a default Mime type defined for .js files so I added one. Not sure if this wil work and if it's the right thing to do even if it does work!
image

dev-badace:
Fixes #328 
 
 In Windows devices, the assetsMap was generating file paths with double backslashes, resulting in incorrect file paths. This PR addresses this issue by replacing the double backslashes with single forward slashes, ensuring that the file paths are correct and functional.

### DIFF
--- a/.changeset/cool-peaches-argue.md
+++ b/.changeset/cool-peaches-argue.md
@@ -1,0 +1,9 @@
+---
+"partykit": patch
+---
+
+[fix] filePath in windows
+
+Fixes #328
+
+In Windows devices, the assetsMap was generating file paths with double backslashes, resulting in incorrect file paths. This PR addresses this issue by replacing the double backslashes with single forward slashes, ensuring that the file paths are correct and functional.
--- a/packages/partykit/src/cli.tsx
+++ b/packages/partykit/src/cli.tsx
@@ -70,7 +70,10 @@ function* findAllFiles(
         }
         dirs.push(filePath);
       } else {
-        yield path.relative(root, filePath);
+        yield path.relative(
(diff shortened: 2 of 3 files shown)


threepointone:
Let's land this and see whether it works

PR #412, opened 2023-09-24
Use existing indentation of package.json during partykit init

tsriram:
Currently when you run `npx partykit@latest init` in an existing project and choose to add partykit to existing `package.json`, we add dependencies to the JSON content and write it back to `package.json` file with 2 spaces here: https://github.com/partykit/partykit/blob/a01be0da91b0b8b18cc50a63bb2a655b08178455/packages/partykit/src/cli.tsx#L217

This is a problem when the existing `package.json` uses a different indentation like 4 spaces / tabs since it's a bit confusing to see a huge `git diff` after the command. Ran into this when running `npx partykit@latest init` in a SvelteKit project since it uses tabs for formatting by default. While one can run re-format the file to see the actual diff, I think it'll be a good experience to keep the existing indentation. And @sindresorhus has a module just for this 😄 

I tried this code locally and everything looks to be working fine.

### DIFF
--- a/packages/partykit/package.json
+++ b/packages/partykit/package.json
@@ -44,6 +44,7 @@
   "dependencies": {
     "@cloudflare/workers-types": "4.20230914.0",
     "clipboardy": "3.0.0",
+    "detect-indent": "^7.0.1",
     "esbuild": "0.19.3",
     "miniflare": "3.20230918.0",
     "yoga-wasm-web": "0.3.3"
--- a/packages/partykit/package.json
+++ b/packages/partykit/package.json
@@ -44,7 +44,6 @@
   "dependencies": {
     "@cloudflare/workers-types": "4.20230914.0",
     "clipboardy": "3.0.0",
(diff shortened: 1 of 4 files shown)


packages/partykit/package.json
```diff
@@ -44,6 +44,7 @@
   "dependencies": {
     "@cloudflare/workers-types": "4.20230914.0",
     "clipboardy": "3.0.0",
+    "detect-indent": "^7.0.1",
```
threepointone, line 47: Could you move this to devDependencies? We bundle our javascript, so it's not needed to be installed by consumers

threepointone (commented):
Great PR! Could you add a patch changeset by running "npx changeset" in the root?

tsriram:
> Great PR! Could you add a patch changeset by running "npx changeset" in the root?

Sure, done. This is the first time I'm adding changeset so let me know if something needs to be updated.

threepointone (approved):
Looks perfect, thank you very much!

PR #416, opened 2023-09-25
blog: Show 10 posts per page

threepointone:
We currently only show 4 (because that's the default setting.) Let's show more!


jevakallio (approved):
LGTM from code perspective!

sylwiavargas (approved):
Good catch!
Also, look at us writing so much!!

PR #417, opened 2023-09-25
Make it more clear in docs how host name is built

patcon:
Presumably this is what comes from partykit.json. If dots are possible, it at least makes the docs slightly more confusing, because it's ambiguous if the host is something like `party.my-project.party.dev`


threepointone:
Good point! I made a small correction and added it to usePartySocket as well. Could you sign the CLA, and I'll land this?

patcon:
Haha oh -- seems I still didn't understand :) This feels way more self-documenting. Thanks!

PR #422, opened 2023-09-26
fixed typo in docs for PartySocket (Client Api)

dev-badace:
lol I thought there was a bug in the partySocket lib when I was seeing the log `connected` even when the servers were not running, turns out it was a typo in the docs from where I copy/pasted the code. Fixed that


threepointone:
oh lol good catch, thank you!

PR #433, opened 2023-09-27
Update example for getting all the rooms messages onStart

Ankcorn:
I think this might be a typo, it tripped me up when I started persisting data :)


threepointone:
whoops, nice catch!

PR #437, opened 2023-09-28
Ensure that user attachments don't clobber system attachments used by PartyKit

### LINKED ISSUE #434: Partykit ceases to function after leaving a tab in the background for a prolonged period
Hello,

I am currently exploring the capabilities of PartyKit and am testing its resilience with respect to leaving a tab in the background and then returning to the app. Unfortunately, my initial experience has not been positive. Below are the steps to reproduce the issue:

1. Visit the [example reactions page on GitHub](https://github.com/partykit/example-reactions).
2. Open the demo app (in my case with a new ID at [this link](https://example-reactions.jevakallio.partykit.dev/test)).
3. Interact with the reactions; everything works as expected.
4. Leave the tab in the background for a while (approximately 2 minutes).
5. Return to the tab and attempt to add a new reaction.

Upon following these steps, I observed that nothing happens when trying to add a new reaction. Although a new message is sent via WebSocket, the object with reactions is not updated as expected.

Zrzut ekranu 2023-09-27 o 22 33 56

https://github.com/partykit/partykit/assets/16654786/80ca7357-d5f6-49b8-9571-84479e41c138

jevakallio:
Fixes #434 
Fixes root cause for performance issue https://github.com/partykit/sketch-voronoi/pull/4

## Problem 

1. In hibernation mode, PartyKit stores basic connection information such as connection id and url in the websocket attachments, so that we can regain access to them when the object wakes up from hibernation.
2. Any user invocation of`serializeAttachment` overwrote the fields internally managed by PartyKit
3. When the object next wakes up in reaction to a message, we assert that the information we store is still present. This assertion fails, and causes a silent error.

## Fix

We "namespace' partykit fields and user fields, and patch the `serializeAttachment` and `deserializeAttachment` fields to return only the value from the user namespace. Internally, we use the getters for the private fields, so no changes to internal platform implementation required.

This is a backwards-compatible change. 

There are a couple of subtle non-breaking behavioural changes, which I'll note inline in the review.

### DIFF
--- a/packages/partykit/facade/connection.ts
+++ b/packages/partykit/facade/connection.ts
@@ -15,42 +15,79 @@ if (!("OPEN" in WebSocket)) {
   Object.assign(WebSocket.prototype, WebSocketStatus);
 }
 
-/** Fields that are stored and rehydrates when a durable object hibernates */
-type ConnectionFields = {
-  id: string;
-  uri: string;
+/**
+ * Store both platform attachments and user attachments in different namespaces
+ */
+type ConnectionAttachments = {
+  __pk: {
+    id: string;
+    uri: string;
+  };
+  __user?: unknown;
 };
 
 /**
  * Cache websocket attachments to avoid having to rehydrate them on every property access.
  */
-const attachments = new WeakMap<WebSocket, ConnectionFields>();
-function deserializeAttachment(ws: WebSocket): ConnectionFields {
-  let attachment = attachments.get(ws);
-  if (!attachment) {
-    // pick only known fields to avoid keeping user's attachments in memory
-    const { id, uri } = ws.deserializeAttachment() as ConnectionFields;
-    attachment = { id, uri };
-    attachments.set(ws, attachment);
+class AttachmentCache {
+  _cache = new WeakMap<WebSocket, ConnectionAttachments>();
+
+  get(ws: WebSocket): ConnectionAttachments {
+    let attachment = this._cache.get(ws);
+    if (!attachment) {
+      attachment = WebSocket.prototype.deserializeAttachment.call(ws);
+      if (attachment !== undefined) {
+        this._cache.set(ws, attachment);
+      } else {
+        throw new Error(
+          "Missing websocket attachment. This is most likely an issue in PartyKit, please open an issue at https://github.com/partykit/partykit/issues"
+        );
+      }
+    }
+
+    return attachment;
   }
 
-  return attachment;
+  set(ws: WebSocket, attachment: ConnectionAttachments) {
+    this._cache.set(ws, attachment);
+    WebSocket.prototype.serializeAttachment.call(ws, attachment);
+  }
 }
 
+const attachments = new AttachmentCache();
+
+const isPatched = (ws: WebSocket): ws is Party.Connection => {
+  return "__is_patched" in ws;
+};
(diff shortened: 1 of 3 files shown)


packages/partykit/facade/connection.ts
```diff
@@ -15,43 +15,80 @@ if (!("OPEN" in WebSocket)) {
...
  */
-const attachments = new WeakMap<WebSocket, ConnectionFields>();
-function deserializeAttachment(ws: WebSocket): ConnectionFields {
-  let attachment = attachments.get(ws);
-  if (!attachment) {
-    // pick only known fields to avoid keeping user's attachments in memory
-    const { id, uri } = ws.deserializeAttachment() as ConnectionFields;
-    attachment = { id, uri };
-    attachments.set(ws, attachment);
+class AttachmentCache {
```
jevakallio, line 32: The big change here is that we now cache both partykit attachments and user attachments. This means that if user uses large attachments, and the have tons of sockets, this may eat into the user's memory budget.

We can implement a smarter caching strategy here, but for now this is preferable, because the perf cost of calling into the platform was already eating into the CPU budget, causing https://github.com/partykit/sketch-voronoi/pull/4

packages/partykit/facade/connection.ts
```diff
@@ -15,43 +15,80 @@ if (!("OPEN" in WebSocket)) {
...
-  return attachment;
+  set(ws: WebSocket, attachment: ConnectionAttachments) {
+    this._cache.set(ws, attachment);
+    WebSocket.prototype.serializeAttachment.call(ws, attachment);
+  }
 }
 
+const attachments = new AttachmentCache();
+
+const isPatched = (ws: WebSocket): ws is Party.Connection => {
```
jevakallio, line 59: Ugly but is it too ugly?

packages/partykit/src/server.ts
```diff
@@ -78,6 +78,11 @@ export type Connection = WebSocket & {
   // It's also set as readonly, so we can't set it ourselves.
   // Instead, we'll use the `uri` property.
   uri: string;
+
+  // Underlying cloudflare types specify the type as any, but unknown is safer
+  // as it forces a cast, so let's add a optional generic overload
+  serializeAttachment<T = unknown>(attachment: T): void;
```
jevakallio, line 84: This is an unrelated change to improve type safety of these APIs. The original issue was unnecessarily hard to debug, because the `any` type swallowed what should have been an error on null property access.

The generic type here is optional, similar to how `storage.get` already works.

packages/partykit/facade/connection.ts
```diff
@@ -15,43 +15,80 @@ if (!("OPEN" in WebSocket)) {
...
+  _cache = new WeakMap<WebSocket, ConnectionAttachments>();
+
+  get(ws: WebSocket): ConnectionAttachments {
+    let attachment = this._cache.get(ws);
+    if (!attachment) {
+      attachment = WebSocket.prototype.deserializeAttachment.call(ws);
+      if (attachment !== undefined) {
+        this._cache.set(ws, attachment);
+      } else {
+        throw new Error(
```
jevakallio, line 42: This should never happen, but showing an error to the user feels safer than just ignoring it.

threepointone (approved):

PR #438, opened 2023-09-28
Add new `connection.state` / `connection.setState` API

jevakallio:
## Proposal

I'm proposing the following API to replace the current serialize/deserializeAttachment API:

```ts
type State = { foo: string, baz?: number } ;

onConnect(conn: Party.Connection<State>) {
  // set state
  conn.setState({ foo: "bar"  });
 
  // read state
  console.log(conn.state); // { foo: "bar" }

  // update state
  const updatedState = conn.setState(prevState => ({ ...prevState, baz: 1 });;
  console.log(updatedState) // { foo: "bar", baz: 1 };
}
```

A more complete example of what this change would look like in user code: https://github.com/partykit/sketch-voronoi/pull/5/files

The same API works for hibernating and non-hibernating connections, and provides optional type safety by adding a generic parameter on `Connection`.

## TODO

- [x] Basic implementation
- [x] Decide on final API
- [x] Implement in platform
- [x] Fix TODOs

## Rationale

In the last week, we've ran into a number of issues with the WebSocket attachment API. The previous PR in this stack (https://github.com/partykit/partykit/pull/437) fixes some of the correctness and performance issues, but still leaves more to be desired:

1. When not in hibernation mode, you can keep state by assigning it to the `connection` object:

```tsx
onConnect(conn: Party.Connection) {
  conn.foo = "bar";
}
```

2. However, if you’re using TypeScript, this will cause a type error, for which you need to do a manual upcast, which is not obvious and feels hacky:

```tsx
onConnect(conn: Party.Connection & { foo?: string }) {
  conn.foo = "bar";
}
```

3. This works, but when you enter hibernation mode, this stops working as soon as the socket hibernates, so you need to serialize the attachment:

```tsx
onConnect(conn: Party.Connection & { foo?: string }) {
  conn.serializeAttachment({ foo: "bar" });
  //later...
  const data = conn.deserializeAttachment();
  console.log(data.foo);
}
```

4. While `serializeAttachment` and `deserializeAttachment` work without hibernation mode, they are unintuitive and don't make sense when the connection stays in memory.
5. `serializeAttachment` is not additive — a socket can only have one attachment, so when you want to edit it, you need to do an awkward read/write:
```ts
  conn.serializeAttachment({ 
    ...(conn.deserializeAttachment()),
    foo: "bar" 
   });
```

### DIFF
--- a/packages/partykit/facade/connection.ts
+++ b/packages/partykit/facade/connection.ts
@@ -81,6 +81,25 @@ export const createLazyConnection = (
     get socket() {
       return ws;
     },
+    get state() {
+      // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+      return this.deserializeAttachment() as Party.ConnectionState<unknown>;
+    },
+
+    setState<T>(setState: T | Party.ConnectionSetStateFn<T>) {
+      let state: T;
+      if (setState instanceof Function) {
+        state = setState(this.state as Party.ConnectionState<T>);
+      } else {
+        state = setState;
+      }
+
+      // TODO: deepFreeze object?
+
+      this.serializeAttachment(state);
+      return state as Party.ConnectionState<T>;
+    },
+
     deserializeAttachment<T = unknown>() {
       const attachment = attachments.get(ws);
       return (attachment.__user ?? null) as T;
@@ -130,7 +149,7 @@ export interface ConnectionManager {
   getCount(): number;
   getConnection(id: string): Party.Connection | undefined;
   getConnections(tag?: string): IterableIterator<Party.Connection>;
-  accept(connection: Party.Connection, tags: string[]): void;
+  accept(connection: Party.Connection, tags: string[]): Party.Connection;
 
   // This can be removed when Party.connections is removed
   legacy_getConnectionMap(): Map<string, Party.Connection>;
@@ -171,6 +190,7 @@ export class InMemoryConnectionManager implements ConnectionManager {
   }
 
   accept(connection: Party.Connection, tags: string[]): void {
+    connection = createStatefulConnection(connection);
     connection.accept();
 
     this.connections.set(connection.id, connection);
@@ -187,6 +207,8 @@ export class InMemoryConnectionManager implements ConnectionManager {
     };
     connection.addEventListener("close", removeConnection);
     connection.addEventListener("error", removeConnection);
+
+    return connection;
   }
 }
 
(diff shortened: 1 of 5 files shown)


packages/partykit/facade/connection.ts
```diff
@@ -79,6 +82,25 @@ export const createLazyConnection = (ws: WebSocket): Party.Connection => {
...
+        state = setState;
+      }
+
+      // TODO: deepFreeze object?
+
+      this.serializeAttachment(state);
+      return state as Party.ConnectionState<T>;
+    },
+
     deserializeAttachment<T = unknown>() {
```
threepointone, line 104: how do you feel about deprecating serializeAttachment/deserializeAttachment on Connection? So after we eventually remove it, users can't futz with the object we've made

packages/partykit/facade/connection.ts
```diff
@@ -90,6 +112,46 @@ export const createLazyConnection = (ws: WebSocket): Party.Connection => {
...
+  connection: Party.Connection<TState>
+): Party.Connection<TState> => {
+  if (isPatched(connection)) {
+    return connection;
+  }
+
+  let _state: TState;
+
+  return Object.assign(connection, {
+    __patched: true,
```
threepointone, line 134: we could use a WeakMap instead of __patched on the object here

threepointone (commented):
strong approval, been waiting for this for a while.

packages/partykit/facade/connection.ts
jevakallio: I have marked them as deprecated in the types in server.ts for now.

Notably, this PR also stacks on top of #437, which already ensures that the user can't override fields that we use internally.

packages/partykit/facade/connection.ts
jevakallio: Ugh, yes, you're right. I was being lazy, but this is the correct way to go

packages/partykit/facade/connection.ts
```diff
@@ -71,52 +71,97 @@ export const createLazyConnection = (
     return ws;
   }
 
-  const connection = Object.assign(ws, {
```
jevakallio, line 74: TIL `Object.assign` doesn't actually create getters, but sets the value of the getter at time of assignment. This makes perfect sense, but didn't think of it before. The code here used to work because the previous values we were setting were all immutable (`id`, `uri`), or functions.

This broke for dynamic `state`, so moved to `defineProperties`

jevakallio:
@threepointone TypeScript crimes in this commit https://github.com/partykit/partykit/pull/438/commits/600715ed2adfc53ec61d71451b949572c8d243ce

PR #443, opened 2023-09-30
fix link in readme

mayank99:
current link goes to some random twitter account 😅


threepointone:
Haha good catch, thank you!

PR #444, opened 2023-09-30
fix typo in guide

mayank99:
`sender` does not exist in `onConnect`. i assume `connection` is the correct one.


threepointone:
Good catch, thank you!

PR #446, opened 2023-10-01
y-partykit: unstable_onRequest

threepointone:
This adds an `unstable_onRequest` to `y-partykit`. Folks want to use http requests to prerender documents (during SSR, for example), and/or persist with an HTTP request.

This is probably temporary while we rethink y-partykit all together.

Usage:

```ts
import { unstable_onRequest } from "y-partykit";
export default {
  onRequest(req, room) {
    return unstable_onRequest(req, room, {
      /* options */
    });
  },
};
```

// cc @wordware-ai

### DIFF
--- a/packages/y-partykit/src/index.ts
+++ b/packages/y-partykit/src/index.ts
@@ -474,3 +474,18 @@ export async function onConnect(
     }
   }
 }
+
+export async function unstable_onRequest(
+  _req: Request,
+  room: Party.Party,
+  options: YPartyKitOptions
+) {
+  const _doc = await getYDoc(room, options);
+
+  // if it's a POST, should we call options.callback here?
+
+  return new Response(
+    "document as json? or as a buffer?"
+    // we should send the response here
+  );
+}
--- a/.changeset/odd-meals-relate.md
+++ b/.changeset/odd-meals-relate.md
@@ -1,0 +1,22 @@
+---
+"y-partykit": patch
+---
+
+y-partykit: unstable_onRequest
+
+This adds an `unstable_onRequest` to `y-partykit`. Folks want to use http requests to prerender documents (during SSR, for example), and/or persist with an HTTP request.
+
+This is probably temporary while we rethink y-partykit all together.
+
+Usage:
+
+```ts
+import { onRequest } from "y-partykit";
+export default {
+  onRequest(req, room) {
+    return onRequest(req, room, {
+      /* options */
+    });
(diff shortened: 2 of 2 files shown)


packages/y-partykit/src/index.ts
```diff
@@ -474,3 +474,18 @@ export async function onConnect(
...
+  room: Party.Party,
+  options: YPartyKitOptions
+) {
+  const _doc = await getYDoc(room, options);
+
+  // if it's a POST, should we call options.callback here?
+
+  return new Response(
+    "document as json? or as a buffer?"
+    // we should send the response here
```
threepointone, line 489: @wordware-ai what logic are you using to serialise the document here?

wordware-ai: In `onRequest` we return the JSON 

```typescript
async onRequest(request: Party.Request) {
    // When we receive a HTTP request for the document we send the latest JSON back
    console.log("Request");
    const roomStorage = new YPartyKitStorage(this.party.storage);
    const ydoc = await roomStorage.getYDoc(this.party.id);

    if (request.method === "GET") {
      if (ydoc) {
        console.log("Ydoc found in room");
        const json = transformer.fromYdoc(ydoc, "default");
        console.log("JSON", JSON.stringify(json));
        // Ydoc will be empty if room hasn't loaded (i.e. the document is not currently being edited)
        if (json.content?.length) {
          return new Response(JSON.stringify(json));
        }
      }
      // If no ydoc in room, load from db
      const prompt = await this.getPrompt(getDb(this.party), this.party);
      if (!prompt) {
        return error(`Prompt with id ${this.party.id} not found`, 404);
      }
      return json(prompt.json);
    }

    return error("Method not allowed", 405);
  }
  ```

threepointone: Hmm then we should probably have an option to transform the Ydoc

wordware-ai: I'm sure that would be helpful though I'm not sure if there's a ton of complexity around each use case. For reference we're using `@hocuspocus/transformer`
`const transformer = TiptapTransformer.extensions(TiptapExtensions());`

threepointone: Maybe we shouldn't have onRequest then, but just a function(id) that returns the YDoc and lets you do whatever with it

threepointone:
closing this one

PR #447, opened 2023-10-01
y-partykit: add onLoad callback

threepointone:
This adds an `onLoad` callback to `y-partykit`. While we can wait for onConnect to finish before we return, this doesn't give us a handle on the actual YDoc. This adds a callback that we can use to get a handle on the YDoc.

### DIFF
--- a/.changeset/calm-buckets-train.md
+++ b/.changeset/calm-buckets-train.md
@@ -1,0 +1,7 @@
+---
+"y-partykit": patch
+---
+
+y-partykit: add onLoad callback
+
+This adds an `onLoad` callback to `y-partykit`. While we can wait for onConnect to finish before we return, this doesn't give us a handle on the actual YDoc. This adds a callback that we can use to get a handle on the YDoc.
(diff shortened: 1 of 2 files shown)


threepointone:
this might not be as useful if we have https://github.com/partykit/partykit/pull/450

jevakallio (commented):
Is the naming semantically correct? Since we have the `load` function, `onLoad` sounds like the complement to that. 

However, in its current implementation, this function would be called every time a new connection is accepted, so the semantics are more like `onConnected`?

What use case are we supporting here?

It's a little complicated to say when a document is actually "loaded", since depending on the source of truth, it can happen either when:
1. Document is hydrated from storage
2. Document is initialized from `load` callback
3. New empty document is created because it's a truly new room
4. Client syncs state to an empty document because user is not using persistence, and the source of truth exists in the client (or p2p network of clients)

Happy to talk in person, and like #448 and #450, I'm not opposed to landing a change that enables this, just want to understand what we are trying to achieve and why.

jevakallio:
closing as discussed

PR #448, opened 2023-10-01
y-partykit: deprecate `persist: true`

threepointone:
While y-partykit's persistence layer was useful to not need a storage layer, in practice it caused too many issues. Further, it's not really useful when there are deep edit chains across clients. This PR deprecates it's usage, and adds a warning to remove it in the future.

### DIFF
--- a/.changeset/breezy-swans-guess.md
+++ b/.changeset/breezy-swans-guess.md
@@ -1,0 +1,7 @@
+---
+"y-partykit": patch
+---
+
+y-partykit: deprecate `persist: true`
+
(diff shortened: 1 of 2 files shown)


threepointone:
another idea is to use persist:true to save the document to DO storage (instead of every edit)

jevakallio (commented):
I do agree the current approach is a bit of a footgun, as we're not able to do yjs gc on the stored document.

However, I don't think we should deprecate `persist`. We do need built-in storage, and we should make it work more reliably. Storing the document in DO storage (or R2) seems like a good approach.

Perhaps we can do this as part of the broader `y-partykit` rework, and leave the current API be?

jevakallio:
As discussed, closing this in favour of keeping `persist` and reimplementing it in a safer way.

PR #450, opened 2023-10-01
y-partykit: export `getYDoc`

threepointone:
Directly get the `Y.Doc` instance

```ts
import {getYDoc} from 'y-partykit';

// ...

const doc = await getYDoc(room, options)
```

### DIFF
--- a/.changeset/mighty-balloons-lie.md
+++ b/.changeset/mighty-balloons-lie.md
@@ -1,0 +1,15 @@
+---
+"y-partykit": patch
+---
+
+y-partykit: export `getYDoc`
+
+Directly get the `Y.Doc` instance
+
+```ts
+import { getYDoc } from "y-partykit";
+
+// ...
+
+const doc = await getYDoc(room, options);
+```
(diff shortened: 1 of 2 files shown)


wordware-ai:
This looks very helpful. I'm assuming this means `onRequest` will be able to access the same ydoc in memory as any connections?

threepointone:
Yup

jevakallio (commented):
This is possibly confusing, since `options` can be provided both in the `onConnect` call and the `getYDoc` call, and we don't have validate that the options match.

- If the user calls `getYDoc` before the first connection is made (as would always be the case in doing this in `onStart`), then the `options` passed would be applied. 
- If `onConnect` is called first, then those `options` are used instead.

I agree there is a need for accessing the ydoc, but the implementation could be more robust.

If this is a workaround for a specific customer and is needed now, we should at bare minimum:

1. Move the args validation we currently do at top of `onConnect` into `getYDoc` (probably a good place to do it regardless)
2. Add a warning if `options` don't match between calls (probably a good idea to do it anyway, since dynamically changing options for different connections in `onConnect` would lead to undefined behaviour already now).

If we do not have an urgent need for this, similarly to #448, we could tackle this as part of a broader rework.

jevakallio:
@wordware-ai this is now landed as `unstable_getYDoc`. See release notes for important caveats:
https://github.com/partykit/partykit/releases/tag/y-partykit%400.0.7

We will likely be revisiting this API in the future, but in either case we will continue to support equivalent functionality.

PR #458, opened 2023-10-03
Update documentation for Connection.state

jevakallio:
Landing this when we land the new API to stable release.

### DIFF
--- a/apps/docs/src/content/docs/reference/partyserver-api.md
+++ b/apps/docs/src/content/docs/reference/partyserver-api.md
@@ -372,27 +372,37 @@ Uniquely identifies the connection. Usually an automatically generated GUID, but
 
 The original URI of the connection request.
 
+
+##### Party.Connection.setState
+
+`setState` allows you to store small pieces of data on each connection. 
+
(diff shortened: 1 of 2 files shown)


apps/docs/src/content/docs/reference/partyserver-api.md
```diff
@@ -372,27 +372,37 @@ Uniquely identifies the connection. Usually an automatically generated GUID, but
...
+Unlike  [`Party.storage`](#partystorage), connection state is not persisted, and will only exist for the lifetime of the WebSocket connection.
 
 ```ts
-connection.serializeAttachment({ username: "jani" });
+connection.setState({ username: "jani" });
 ```
 
-**Related guide:** [Scaling PartyKit Servers with Hibernation](/guides/scaling-partykit-servers-with-hibernation)
+:::danger[State size]
+The maximum size of the state you can store on a connection is 2KB. If you try to store values larger than 2KB, `setState` will throw an error.
```
sylwiavargas, line 387: could you include the error? this will be helpful in case someone is just searching for the error they got and are confused.

jevakallio: Good idea! I'll add that as an enhancement afterwards

PR #464, opened 2023-10-05
Fix broken link in docs

tsriram:
Link to examples from `Responding to HTTP requests` guide page looks to be broken. Fixing it.


threepointone (approved):
Whoops nice catch, thank you!

PR #467, opened 2023-10-06
Allow passing PartySocket.query as a function

jevakallio:
This enables people to pass dynamic parameters such as authentication tokens upon reconnect.

As a side effect, we can no longer provide a static party `url`, so this is now available as `roomUrl`. `url` now reverts to the original base class behaviour, where it returns the dynamic url of the websocket (or an empty string if websocket is not yet initialized).

This is a breaking change, but AFAIK our only user of this behaviour is @jamesgpearce.

### DIFF
--- a/packages/partykit/src/server.ts
+++ b/packages/partykit/src/server.ts
@@ -148,7 +148,7 @@ export type Party = {
   ) => void;
 
   /** Get a connection by connection id */
-  getConnection<TState = unknown>(id: string): Connection<TState>;
+  getConnection<TState = unknown>(id: string): Connection<TState> | undefined;
 
   /**
    * Get all connections. Optionally, you can provide a tag to filter returned connections.
--- a/packages/partysocket/src/index.ts
+++ b/packages/partysocket/src/index.ts
@@ -1,6 +1,7 @@
 import type * as RWS from "./ws";
 import ReconnectingWebSocket from "./ws";
 
+type Params = Record<string, string>;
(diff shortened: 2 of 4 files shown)


packages/partykit/src/server.ts
```diff
@@ -148,7 +148,7 @@ export type Party = {
   ) => void;
 
   /** Get a connection by connection id */
-  getConnection<TState = unknown>(id: string): Connection<TState>;
+  getConnection<TState = unknown>(id: string): Connection<TState> | undefined;
```
jevakallio, line 151: Fixes accidental omission from #466

threepointone (approved):

sylwiavargas (approved):
thank you!

packages/partysocket/src/index.ts
```diff
@@ -65,34 +66,46 @@ export default class PartySocket extends ReconnectingWebSocket {
...
-      url += `?${new URLSearchParams({ ...query, _pk }).toString()}`;
-    } else {
-      url += `?_pk=${_pk}`;
-    }
 
-    super(url, protocols, socketOptions);
+    const makeUrl = (query?: Params) =>
+      `${baseUrl}?${new URLSearchParams({ ...query, _pk }).toString()}`;
+
+    // allow urls to be
```
jevakallio, line 85:
```suggestion
    // allow urls to be defined as functions
```

PR #468, opened 2023-10-06
Allow passing a function to YPartyKitProvider options.params

jevakallio:
Useful for e.g. passing authentication token and refreshing them on reconnect.


sylwiavargas (approved):
thank you! appreciate it!

PR #470, opened 2023-10-07
added missing crypto from node_compat

dev-badace:
crypto was missing from the node_compat, and users can't use the node:crypto.

I've added it, & it works. well it works on my device 😂


threepointone (approved):
LGTM, thank you!

Issue #471, opened 2023-10-07
`npx partykit deploy` fails


nichoth:
I'm trying `npx partykit deploy` for the first time. It opens github in a web browser, but then I click 'authorize' and nothing happens. In the browser console are some errors 

![image](https://github.com/partykit/partykit/assets/5776508/e6c7a507-b6dc-4a75-af90-3ab214756458)

![image](https://github.com/partykit/partykit/assets/5776508/1b4a8dfc-b5e5-4aee-af38-133f4e1663f6)

> In the meanwhile, you can use the legacy login mechanism which bypasses the PartyKit dashboard and logs you in directly via GitHub:
```
npx partykit login -p github
```

jevakallio:
@nichoth thanks for reporting this! I've fixed this and released it as part of [`partykit@0.0.29`](https://github.com/partykit/partykit/releases/tag/partykit%400.0.29).

It seems that Chrome has started to roll out Private Network Access in Chrome 117. Curiously, upgrading to 117 did not break this for me, but enabling "Respect the result of Private Network Access preflights" in `chrome://flags` did, and I was able to fix the issue.

jevakallio:
Somewhat more concerningly, the login callback seems to be broken on Safari and Brave currently, due to WebKit/Safari choosing not to observe the W3C Mixed Content specification and blocking localhost requests from public websites despite the spec explicitly.

https://bugs.webkit.org/show_bug.cgi?id=171934

I'll keep this issue open until we have a fix in place for WebKit.

jevakallio:
Fixed by #473

PR #473, opened 2023-10-09
Fix login on WebKit

jevakallio:
Fixes remaining issue on #471 for Safari/Brave.

## Problem

WebKit chooses not to observe the W3C Mixed Content specification and blocking localhost requests from public `https` websites despite the spec explicitly allowing it. See: https://bugs.webkit.org/show_bug.cgi?id=171934

## Solution

This PR adds a workaround to perform authentication as a navigation event if the initial callback fails.

This is how `wrangler` does it, as well.

### DIFF
--- a/.changeset/tricky-teachers-occur.md
+++ b/.changeset/tricky-teachers-occur.md
@@ -1,0 +1,5 @@
+---
+"partykit": patch
+---
+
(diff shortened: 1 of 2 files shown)


threepointone (approved):
great catch, dope fix, thank you!

PR #476, opened 2023-10-10
await multiparties.socket()

threepointone:
This introduces a new method `.socket()` to the multiparties bag. Much like the .fetch() method() on the bag, this now uses an internal api to connect with a websocket to the party, instead of using a URL and connecting via the internet. Importantly, this also works from inside onFetch() handlers. This patch also deprecates the previous `.connect()` method.


jevakallio (approved):
Great find!

I'm also happy on the renaming, I think `connect()` was ambiguous in the first place.

I think this makes it impossible to use PartySocket (https://github.com/partykit/partykit/pull/395) for reconnection, buffering etc. We should (eventually) figure out that story for the intra-party sockets, but no reason not to block on this.

PR #478, opened 2023-10-10
partysocket: mark react as an optional dependency

### LINKED ISSUE #477: partysocket: react as a peer dependency?
I've been trying to integrate partysocket in my app, but it messes up my build, installs and uses its own version of react because it's specified as a `dependency` in the package.json:

https://github.com/partykit/partykit/blob/73dac9473ac26d776ad0d7c150dcf3c7a3a195b8/packages/partysocket/package.json#L43

Would you consider changing react from a dependency into a peerDependency? I believe it is [the recommended way](https://stackoverflow.com/a/30454133) for reusable code and libraries.

Happy to make a PR if needed

marekhrabe:
Fixes #477 

If there were some tests for the react hook part of the partysocket, it might make sense to add it also as a `devDependency`, but it wasn't currently needed for anything.


threepointone (approved):
Looks great, thank you very much for adding a changeset and signing the CLA! This will go out in the next release (but you should be able to use it with the `@beta` tag soon after it lands)

alitnk:
FWIW this breaks my Expo app with "Unable to resolve "partysocket/react" from "file.tsx" - reverting to `0.0.9` fixed it.

PR #483, opened 2023-10-11
Improve y-partykit storage 

jevakallio:
This PR starts work towards improving the robustness and scalability of `y-partykit`. 

## Problem

Currently, the y-partykit storage adapter stores every update in room storage, as well as the entire document state encoded as an update when the last user leaves the room. This means the room only grows, and for large documents, can grow rather quickly.

There was logic in place for collapsing the update log when it gets too long (currently over 300 entries), but this was not working due to a bug in the `clearRange` routine.

## Change

When room becomes empty, instead of appending document state to update log, merge current update log into a single state updated.

## Scope

This may be sufficient to prevent unbounded growth of room storage.

This may not solve issues for existing users experiencing crashes where the unpacked room storage size exceeds the Durable Object memory limit -- for those, we may need to implement a pre-check that reads keys from storage in smaller chunks and cleans up unnecessary duplicated data before allowing a full range read for the entire update log.


threepointone (approved):
code looks good to me!

PR #484, opened 2023-10-11
[PROPOSAL] y-partykit storage modes

jevakallio:
Adds options to control how to persist the document to PartyKit room storage.

The goal here is to not break existing clients, and provide a migration path to more sensible strategies.

 - `{mode: "snapshot"}` — persist full document snapshot (recommended)
 - `{mode: "history", maxUpdates, maxBytes }` — persist document edit history
 - `true` — Equivalent to `{ mode: "history" }` (deprecated, use `{ mode: "history "}` instead)
 - `false` — Do not persist document or history (default value)

See diff for better options.

### DIFF
--- a/packages/y-partykit/src/index.ts
+++ b/packages/y-partykit/src/index.ts
@@ -14,6 +14,8 @@ function assert(condition: unknown, message: string): asserts condition {
   }
 }
 
+const MAX_PERSIST_BYTES = 10_000_000;
+
 const wsReadyStateConnecting = 0;
 const wsReadyStateOpen = 1;
 const wsReadyStateClosing = 2; // eslint-disable-line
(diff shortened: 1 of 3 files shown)


packages/y-partykit/src/index.ts
```diff
@@ -429,12 +457,47 @@ interface UrlCallbackOptions extends CallbackOptions {
...
+       * Keeps full document history as long as there are connected clients,
+       * and compresses changes to a snapshot when last client disconnects.
+       **/
+      mode: "snapshot";
+    };
+
 export type YPartyKitOptions = {
   /**
-   * disable gc when using snapshots!
+   * disable gc when using persist or persistStrategy!
```
jevakallio, line 486:
```suggestion
   * disable gc when using persist!
```

PR #494, opened 2023-10-12
Support Y.js messages > 1MB

jevakallio:
## Problem

The Workers platform rejects WebSocket messages when they are above 1MB.

The Y.js provider does not like this, and end up in a nasty reconnect/retry loop, which will spam the room with very large messages.

## Solution

Chunk large Y.js update messages to 1MB chunks and reassemble them on the server side.

This implementation is not 100% resilient to network issues such as dropped messages, but I've tried to validate the inputs as well as I can and not pass malformed messages to the handler on the server. In this case, the worst case behaviour is the same as it was before applying this fix in the first place.

## Considerations

Generally, limiting WebSocket message sizes is a good idea, but with Y.js it's reasonable for the document size to grow to exceed 1MB.  Sending the large message should only happen on initial sync, when the server doesn't have a state vector for the document.

### DIFF
--- a/packages/y-partykit/src/chunking.ts
+++ b/packages/y-partykit/src/chunking.ts
@@ -1,0 +1,172 @@
+// This file contains a shared implmentation of chunking logic for binary
+// WebSocket messages. Because the PartyKit platform limits individual
+// WebSocket messages to 1MB, we need to split larger messages into chunks.
+export const CHUNK_MAX_SIZE = 1_000_000;
+
+const BATCH_SENTINEL = "y-pk-batch";
+const TRACE_ENABLED = false;
+const trace = (...args: unknown[]) => TRACE_ENABLED && console.log(...args);
+
+let warnedAboutLargeMessage = false;
+
+type MessageData = ArrayBufferLike | string;
+type MessageHandler = (message: { data: MessageData }) => void;
+type Batch = {
+  id: string;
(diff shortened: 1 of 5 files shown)


packages/y-partykit/src/chunking.ts
```diff
@@ -0,0 +1,172 @@
+// This file contains a shared implmentation of chunking logic for binary
```
jevakallio, line 1:
```suggestion
// This file contains a shared implementation of chunking logic for binary
```

threepointone (approved):
Love reading your PRs. No notes.

dev-badace:
an alternative solution would be to send messages over 1mb over http.

packages/y-partykit/src/chunking.ts
```diff
@@ -0,0 +1,179 @@
+// This file contains a shared implementation of chunking logic for binary
+// WebSocket messages. Because the PartyKit platform limits individual
+// WebSocket messages to 1MB, we need to split larger messages into chunks.
+export const CHUNK_MAX_SIZE = 1_000_000;
```
BrianHung, line 4: Should this be 1_048_576? Was reading cloudflare docs and noticed they do 1 MiB instead of 1 MB: https://developers.cloudflare.com/durable-objects/platform/limits/

PR #496, opened 2023-10-13
update deps

threepointone:
The big one to test here is the clerk update. My local setup is a bit broken so I can't test that right now


threepointone:
I reverted the clerk update so we can land the rest

PR #499, opened 2023-10-14
partysocket: add subpath support

threepointone:
This patch lets you add `path: string` to `new PartySocket({...})` and point to a subpath in a room. This is a client side analog to the recent subpath support we added to multiparty `.fetch()`/`.socket()`

### DIFF
--- a/packages/partysocket/src/index.ts
+++ b/packages/partysocket/src/index.ts
@@ -15,6 +15,7 @@ export type PartySocketOptions = Omit<RWS.Options, "constructor"> & {
   party?: string; // the party to connect to (defaults to main)
   protocol?: string;
   protocols?: string[];
+  path?: string;
   query?: Params | (() => Params | Promise<Params>);
   // headers
 };
@@ -58,6 +59,7 @@ export default class PartySocket extends ReconnectingWebSocket {
   name: string;
   room: string;
   host: string;
+  path: string;
 
   constructor(readonly partySocketOptions: PartySocketOptions) {
     const {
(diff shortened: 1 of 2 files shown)


packages/partysocket/src/index.ts
```diff
@@ -78,12 +81,17 @@ export default class PartySocket extends ReconnectingWebSocket {
       host = host.slice(0, -1);
     }
 
+    if (rawPath && rawPath.startsWith("/")) {
```
threepointone, line 84: hmm maybe it _should_?

jevakallio: Aesthetically, it probably should, but then again it's really NOT a root path since it's relative to the party url, so I think just `"path/foo"` is more correct

jevakallio (commented):
Code-wise, this is looking good.

However, we still seem to have diverging implementations when using the main party default url `/party/:id` vs named party address (`/parties/:party/:id`).

Specifically, the following produces the id `some-room`:
```ts
const partySocket = new PartySocket({
  host: PARTYKIT_HOST,
  room: "some-room",
  party: "main",
  path: "subpath",
});
```


But the following produces the id `some-room/subpath`:
```diff
const partySocket = new PartySocket({
  host: PARTYKIT_HOST,
  room: "some-room",
-  party: "main",
  path: "subpath",
});
```

We should probably fix this before landing/communicating the new path feature to the users.

Tested with:
```
  static onBeforeConnect(req: Party.Request, lobby: Party.Lobby) {
    console.log("lobby.id", lobby.id);
    console.log("req.url", req.url);
    return req;
  }
```

threepointone:
Fix for the id thing in https://github.com/partykit/partykit/pull/503

PR #500, opened 2023-10-14
DO NOT LAND partykit: adds more hibernatable socket methods

threepointone:
This adds the missing methods that we can use to respond to messages when objects are hibernated without waking up the object. Namely, setWebSocketAutoResponse, getWebSocketAutoResponse, and getWebSocketAutoResponseTimestamp. Details in type signature, more info here: https://developers.cloudflare.com/durable-objects/api/websockets/#setwebsocketautoresponse

### DIFF
--- a/.changeset/orange-eels-happen.md
+++ b/.changeset/orange-eels-happen.md
@@ -1,0 +1,7 @@
+---
+"partykit": patch
+---
+
+partykit: adds more hibernatable socket methods
+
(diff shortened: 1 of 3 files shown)


jevakallio:
Dropping a quick note based on IRL chat, we should add some default partykit metadata on the response, so that we can use this as a room-level meta storage for room ids, connection counts, etc.

PR #503, opened 2023-10-17
fix: get proper room id from `/party/:id/*`

threepointone:
We had a bug when picking out the room name from `/party/:id/*`, it would pick the whole subpath instead. This fixes it.


threepointone:
This is ostensibly a breaking change, so let's mention on the discord?

PR #509, opened 2023-10-17
Parallellize asset uploads + fail fast

jevakallio:
Runs asset uploads in parallel. Adds a pre-upload step to validate assets before attempting to upload, so user doesn't have to wait until end of the upload batch before finding there was an error.

### DIFF
--- a/packages/partykit/src/cli.tsx
+++ b/packages/partykit/src/cli.tsx
@@ -497,7 +497,7 @@ export async function deploy(options: {
     fileName: string;
   }[] = [];
 
-  //
+  // prepare static assets to be uploaded
 
   if (assetsPath) {
     // do a build
@@ -570,6 +570,8 @@ export async function deploy(options: {
     }
   }
 
+  // before starting upload, let's build the source files (fail fast)
+
   const wasmModules: Record<string, Buffer> = {};
 
   const code = (
@@ -631,6 +633,8 @@ export const ${name} = ${name}Party;
       ],
     })
   ).outputFiles![0].text;
+
+  // starting upload. first, uploading assets...
 
   if (filesToUpload.length > 0) {
     logger.log(
@@ -697,6 +701,8 @@ export const ${name} = ${name}Party;
     });
   }
 
+  // then, prepare the deploy request
+
   const form = new FormData();
   form.set("code", code);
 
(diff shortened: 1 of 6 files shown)


packages/partykit/src/cli.tsx
```diff
@@ -476,12 +479,27 @@ export async function deploy(options: {
...
-    const params = options?.preview
-      ? `?${new URLSearchParams({ preview: options.preview })}`
-      : "";
-    const assetsApiPath = `/parties/${user.login}/${config.name}/assets${params}`;
+  const assetsApiParams = options?.preview
+    ? `?${new URLSearchParams({
+        preview: options.preview,
+        // notify that the client has attempted to call prepare_assets,
+        // so that each upload process doesn't need to do it
+        prepare: "true",
```
jevakallio, line 487: We are sending an explicit flag to the PUT requests to say that the client has already attempted to call `prepare_assets`. This will reduce CloudFlare API invocations by `n/2` when each upload operation on the platform doesn't need to first check for the bucket existence.

This is kind of hacky, but since we want backwards compatibility for old CLIs, we need to signal this explicitly.

packages/partykit/src/cli.tsx
```diff
@@ -544,42 +568,6 @@ export async function deploy(options: {
         });
       }
     }
-
-    if (filesToUpload.length > 0) {
```
jevakallio, line 548: Moving this to after the esbuild source compilation so in case the compilation fails, we won't have unnecessarily uploaded the new assets that will get orphaned when the deploy never completes.

packages/partykit/src/cli.tsx
```diff
@@ -644,6 +632,71 @@ export const ${name} = ${name}Party;
...
+                  "/"
+                )}`
+              );
+            })
+          )
+        )
+      )
+    );
+
+    logger.log(`Updating asset index...`);
```
jevakallio, line 688: Adding this because when you are deleting a large amount of old assets, the asset map commit step can take a while, which may make it seem like the process has hung.

packages/partykit/src/cli.tsx
```diff
@@ -695,6 +748,8 @@ or by passing it in via the CLI
     urlSearchParams.set("preview", options.preview);
   }
 
+  console.log("Deploying...");
```
jevakallio, line 751: Similarly to above, adding this step to signal when the previous step has completed.

threepointone (approved):

PR #520, opened 2023-10-25
Add OG image generation to docsite

jevakallio:
Adds dynamic OpenGraph images. 

## Implementation

The current version of starlight does not directly support dynamic opengraph images. This PR relies on a separate Next.js site that uses vercel/og to generate images, and patches the starlight package to render the correct meta tags into the page.

Once starlight adds better meta override support, we can remove the patch.

## Examples

image

image

## Live example


sylwiavargas (approved):
Thank you for creating this!
It really helps visibility and shareability of our docs and pages.

PR #521, opened 2023-10-25
update dependencies

threepointone:
Of note, this now uses miniflare's new coloured logging, so we can start deprecating our weird inspector layer


jevakallio (approved):
Ohh dope!

PR #523, opened 2023-10-25
splits Quickstart pages to two, reorders sidebar

sylwiavargas:
This PR:

- refreshes the content on "What is PartyKit"
- fixes an error in the content around running dev server (the experience is not the same for the new app, and when adding PartyKit to an existing app)
- updates the content around deployment
- rearranges the "Get Started" section so folks can faster get going

### DIFF
--- a/apps/docs/src/content/docs/quickstart.md
+++ b/apps/docs/src/content/docs/quickstart.md
@@ -36,7 +36,7 @@ Once the server is running, open localhost with the designated port (defaulting 
 Next, deploy your PartyKit server:
 
 ```bash
-npm partykit deploy
+npx partykit deploy
 ```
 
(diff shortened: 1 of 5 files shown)


apps/docs/src/content/docs/quickstart.md
```diff
@@ -41,15 +33,13 @@ Once the server is running, open localhost with the designated port (defaulting
 
 ## Deploy your app
 
-To deploy your app, run the following command in your terminal in your project's directory:
+Next, deploy your PartyKit server:
 
 ```bash
-npx partykit deploy
+npm partykit deploy
```
jevakallio, line 39: I think this should be `npx`
```suggestion
npx partykit deploy
```

Alternatively, it could be `npm run deploy` since the project does create the alias in package.json, but since we recommend using `npx run dev` above, I'd do the same here.

jevakallio (approved):
Looking great!

PR #526, opened 2023-10-26
Fix broken links 

vimode:
This PR fixed a few broken links.

- In the sand-game, without the https:// for sand.graphics, it became a relative link I guess linking to https://docs.partykit.io/examples/app-examples/falling-sand-game-three-js/sand.graphics. 

- The youtube watch party link 404ed, so updated it to link to the live demo of *Namib Desert, watering hole livestream* example demo.

Thank you.


sylwiavargas (approved):
Awwww thank you!
You're be the hero that we need, but we don't deserve 🥰

PR #527, opened 2023-10-27
Add more guides

jevakallio:
Adds more guides to the docs:
- Alarms
- Rate limiting
- Validating client inputs with Zod
- Binary message formats, and encoding with MessagePack


sylwiavargas (approved):
what a lovely way to end the week 🥳
thank you!

PR #529, opened 2023-10-27
Improvements to usePartySocket (and useWebSocket) reliability

jevakallio:
This PR fixes multiple issues in `usePartySocket` and `useWebSocket`:

- When props such as url, host, party etc were set dynamically, the socket would not reconnect and would remain connected with the existing parameters
- When event handlers were provided as functions, the socket would reconnect when function references changed

This change (seemingly) introduces some additional complexity to the code. In order to avoid duplication of said complexity between `usePartySocket` and `useWebSocket` I introduced some additional indirection, but I think that's worth it in this case.

Ideally we should have automated tests for this, but testing react hook lifecycles is a little tricky and we don't have infra set up for it, so in lieu I extensively tested this manually in different scenarios (strict mode, non-strict, etc) and it seems to work as expected.

### DIFF
--- a/packages/partysocket/src/react.ts
+++ b/packages/partysocket/src/react.ts
@@ -24,6 +24,7 @@ export default function usePartySocket(options: UsePartySocketOptions) {
         options.query,
         options.id,
         options.host,
+        options.room,
         options.party,
         options.path,
         options.protocol,
--- a/packages/partysocket/src/react.ts
+++ b/packages/partysocket/src/react.ts
@@ -1,13 +1,35 @@
-import { useEffect, useRef } from "react";
+import { useEffect, useRef, useMemo, useState } from "react";
 import type { PartySocketOptions } from ".";
 import PartySocket from ".";
 import useWebSocketImpl from "./use-ws";
 
-type UsePartySocketOptions = PartySocketOptions & {
+type EventHandlerOptions = {
   onOpen?: (event: WebSocketEventMap["open"]) => void;
   onMessage?: (event: WebSocketEventMap["message"]) => void;
   onClose?: (event: WebSocketEventMap["close"]) => void;
   onError?: (event: WebSocketEventMap["error"]) => void;
+};
+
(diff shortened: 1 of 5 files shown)


packages/partysocket/src/use-socket.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  // keep track of the socket we initialized
+  const socketInitializedRef = useRef<T | null>(null);
+
+  // allow changing the socket factory without reconnecting
+  const createSocketRef = useRef(createSocket);
+  createSocketRef.current = createSocket;
+
+  // finally, initialize the socket
+  useEffect(() => {
```
jevakallio, line 54: Some disgustingness here, I tried to leave excessive comments as breadcrumbs

packages/partysocket/src/react.ts
```diff
@@ -1,61 +1,43 @@
...
-      };
-    },
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    []
-  );
-  return socketRef.current;
+  const socket = useStableSocket({
+    options,
+    createSocket: (options) => new PartySocket(options),
+    createSocketMemoKey: (options) =>
```
mattrossman, line 20: Could `options.room` be added as a memo key? Found this PR because I was expecting the socket to reconnect on room change.

packages/partysocket/src/react.ts
```diff
@@ -1,61 +1,43 @@
...
+      JSON.stringify([
+        // NOTE: if query is defined as a function, the socket
+        // won't reconnect when you change the function identity
+        options.query,
+        options.id,
+        options.host,
```
jevakallio, line 26:
```suggestion
        options.host,
        options.room,
```

packages/partysocket/src/react.ts
jevakallio: Oh -- great catch, it was indeed supposed to be in this list. Added! 🙌

threepointone (approved):
This looks great, thanks Jani.

PR #530, opened 2023-10-27
Fix stub.fetch types to allow forwarding Request directly

jevakallio:
Makes it possible to route requests without type hackery, as in https://github.com/partykit/example-sharding/blob/main/src/router.ts#L22

```diff
    return lobby.parties.main
      .get(id)
-     .fetch(req as unknown as RequestInit);
+     .fetch(req);
```


threepointone (approved):
lgtm!

PR #535, opened 2023-10-31
Fixed type error

DhanushNehru:
'ą' character
Noticed while reading the documentation. So created a PR for the same to remove it


threepointone:
Good catch, thank you!

threepointone (approved):

PR #539, opened 2023-11-01
validating project name in config

andrew-r-thomas:
I had an issue with my project name previously, I had an underscore in the name of the project, and since the project name is included in the url for the party, I had two separate issues:

1. i could not deploy initially, there was a problem with uploading my files that was caused by the path name
2. i had a ERR_SSL_VERSION_OR_CIPHER_MISMATCH error in the client when trying to connect via partysocket

this second error I think was caused by subdomain issue in cloudflare https://developers.cloudflare.com/ssl/troubleshooting/version-cipher-mismatch/

this pr just adds a simple check to the config file to make sure that the name of the project will work as a url path segment

### DIFF
--- a/.changeset/warm-tables-jump.md
+++ b/.changeset/warm-tables-jump.md
@@ -1,0 +1,5 @@
+---
+"partykit": patch
+---
+
+validating project name in config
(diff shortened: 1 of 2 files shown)


threepointone:
Oh wow, I think you solved it. Thanks for sending a PR with a fix! I'll review and land this later tonight.

threepointone (approved):

Issue #298, opened 2023-08-30
validate roomnames 


threepointone:
in preparation for subpaths

Issue #117, opened 2023-04-01
validation on room names 


threepointone:
either we should allow looser room names (with special characters), or we should do proper validation and error properly on them.

PR #540, opened 2023-11-02
adding room validation to partysocket

andrew-r-thomas:
this adds another very simple check that the room name in the partysocket config is a valid path segment, there might be more options for room names depending on how you guys are implementing things on the back end, but at least now there is a central place to do the validation. Hoping this (at least partially) covers what is meant in issues #298  and #117 

this doesn't address connecting to parties from within a party server, but if pr #395 gets implemented, that issue will be solved

this was a little bit of guess work as to what the goals were here but I hope it helps!

### DIFF
--- a/.changeset/selfish-kids-guess.md
+++ b/.changeset/selfish-kids-guess.md
@@ -1,0 +1,5 @@
+---
+"partysocket": patch
+---
+
+adding room validation to partysocket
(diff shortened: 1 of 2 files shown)


threepointone (approved):
lgtm, thank you! we'll handle inter party stuff later.

threepointone:
whoops I forgot to add this to the release, so it'll go out in the next one.

PR #585, opened 2023-12-01
Blog reskin

ThinkMake:
@threepointone can you have a quick look through the posts to make sure I haven't broken anything (like embeds etc)


threepointone:
I'll keep adding notes here as I see them 

- the buttons on chrome start with black text and change to white on hover 
- the body copy feels a bit small on phones, could we kick it up a bit?

threepointone:
Could we add a treatment to the author name and date inside a post details page? just to mark it as different from the rest of the post

Issue #587, opened 2023-12-02
Change the room on PartySocket


ayoubqrt:
Hi, I want to create the PartySocket object and connect it later with another room set

So the usage will be like:
```ts
const partySocket = new PartySocket({ 
	host: "localhost:1999", 
	room: "some-room", // or without the room set if startClosed is true 
	startClosed: true 
});

partySocket.setRoom("another-room");
partySocket.reconnect();
```

ayoubqrt:
Solved in #588

PR #588, opened 2023-12-02
Change properties of PartySocket after instantiation

ayoubqrt:
Hi,
This PR solves https://github.com/partykit/partykit/issues/587
Initially, I only wanted to change the room of the object but finally why not change all the properties ? 

I didn't find here how to make the compiler happy, I set the internal properties in `setWSProperties` method called in the constructor but the compiler want me to set them directly in the constructor.
https://github.com/partykit/partykit/compare/main...ayoubqrt:partykit:feat/change-room?expand=1#diff-bb59a004a06ba91d40ba4aea4a6a318fd07bf84e83b826f57de223af14d21449R128-R133


threepointone:
I just merged, it should be published in a few minutes, try in 10

ayoubqrt:
I forgot but it will be good to add it on the docs website. I am reading the page of PartySocket to make some updates

PR #707, opened 2024-01-10
Use blockConcurrencyWhile when running onStart()

threepointone:


packages/partykit/facade/source.ts
```diff
@@ -540,7 +540,10 @@ function createDurable(
         ? new HibernatingConnectionManager(this.controller)
         : new InMemoryConnectionManager();
 
-      return this.worker.onStart();
+      await this.controller.blockConcurrencyWhile(async () => {
+        await this.worker!.onStart();
+      });
+      return;
```
Burry, line 546: @threepointone Encountering an error that I suspect is from this change:

```
Error: A call to blockConcurrencyWhile() in a Durable Object waited for too long. The call was canceled and the Durable Object was reset.
```

threepointone:
Ok I'll revert this change

packages/partykit/facade/source.ts
threepointone: Sorry for the trouble here. This is reverted and published in `partykit@0.0.69`.

PR #723, opened 2024-01-18
create: use templates from partykit/templates

threepointone:
https://github.com/partykit/templates


packages/create-partykit/src/index.tsx
```diff
@@ -142,49 +109,31 @@ async function initGit({ cwd }: { cwd: string }) {
   }
 }
 
+const templateChoices = {
```
threepointone, line 112: @genmon 👇

PR #764, opened 2024-01-28
when singlePageMode is true, for missing urls only return index.html …

threepointone:
…if Sec-Fetch-Mode === 'navigate'

### DIFF
--- a/.changeset/healthy-llamas-eat.md
+++ b/.changeset/healthy-llamas-eat.md
@@ -1,0 +1,5 @@
+---
+"partykit": patch
+---
+
+when singlePageMode is true, for missing urls only return index.html if Sec-Fetch-Mode === 'navigate'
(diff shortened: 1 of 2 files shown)


patcon:
Hi! This means that this won't work to access singlePageApp on LAN like http://192.168.6.129:1999

```
$ partykit dev --live
```

And so `--https` flag is required to access the dev app on the local network:

```
$ partykit dev --live --https
```

Is this desired? I've been using the dev server to run fun local-first apps at events, and the UX of these fun little apps running on the local network, takes a hit over SSL.

Could we do what vite and webpack-dev-server do (via ) and just check for `Accept: text/html` as the way to detect?

https://github.com/bripkens/connect-history-api-fallback/

> Fallback to index.html for applications that are using the HTML 5 history API

https://github.com/bripkens/connect-history-api-fallback/blob/main/lib/index.js#L104

https://github.com/webpack/webpack-dev-server/blob/main/package.json#L60

https://github.com/vitejs/vite/blob/d64a1a5557b3caea9469e70b647ff2c9d9def809/packages/vite/src/node/server/middlewares/htmlFallback.ts#L32

PR #809, opened 2024-02-19
Improved Docs

Itsnotaka:
### **Breaking (Possible)**- Switched npm worksapce to pnpm workspace 
- Made tailwind config use typescript and installed missing `@astrojs/starlight-tailwind` dependency
- Rewrote and added some manual install instructions to [Add installation instructions and manual setup guide for PartyKit](https://github.com/partykit/partykit/commit/a740d88200e33838431003b1f031bf67fa91cea1)
- Monorepo lint using [sherif](https://github.com/QuiiBz/sherif)


threepointone:
We picked npm to keep it simpler for folks who might want to contribute, it's not too bad. If we had to update, we'd probably pick bun once it's ready for windows. 

sherif looks great, we should add that. 

I recommend putting the docs changes in a different PR, and the sherif change in a different PR. 

The lint failures are because of something weird happening with starlight, I'll have a look at that today.

Thanks so much @Itsnotaka!

PR #839, opened 2024-03-05
chore: Updated BlockNote example

matthewlipski:
Hi, we recently updated BlockNote to v0.12.0 which introduced a number of small breaking changes. This PR updates the example to use the new version.


threepointone:
Thanks very much!

PR #844, opened 2024-03-16
y-partykit provider can generate IDs even if 'crypto' is not defined

OCA99:
y-partykit was failing in my environment (VSCode extension running in development mode) because 'crypto' is not defined. Simply adding an extra check to the already existing logic in order to fall back to insecure ID generation when 'crypto' is not defined.


threepointone (approved):
Great catch, thank you!

PR #857, opened 2024-04-01
docs: change room to party

lawrencecchen:


threepointone:
great catch! can you rename this.party to this.room instead?

threepointone:
Thanks!

PR #906, opened 2024-08-18
Move y-partykit deps to peerDependencies

alexdlm:
Marks yjs and react as peerDependencies.

The react hook (and dependency) would ideally be in a separate package, but I've made it optional here.


threepointone:
This seems reasonable. I can't remember why I'd put it directly in deps, probably a mistake. I'll test this next week and land it. Thank you!

threepointone:
I don't think this should be a breaking change, but let's ship it and see what happens.

PR #914, opened 2024-09-19
Include hibernation info

third774:
Hey there! 👋🏻 I was pretty sure `onStart` runs when waking up from hibernation (and according [to this](https://github.com/threepointone/partyserver/blob/main/packages/partyserver/README.md#lifecycle-hooks) it does) so figured I should open a PR to update the docs here as well!


threepointone (approved):
thank you!

PR #928, opened 2024-12-21
Nodejs compat upgrade

maryam-khan-dev:
## Summary
Since September 23, 2024, Cloudflare [has allowed devs to opt into improved Node.js compatibility](https://blog.cloudflare.com/more-npm-packages-on-cloudflare-workers-combining-polyfills-and-native-code/). This update enables Wrangler users to work with [many more Node.js APIs](https://workers-nodejs-compat-matrix.pages.dev/) than is [currently available](https://github.com/partykit/partykit/blob/main/packages/partykit/src/nodejs-compat.ts).

Behind the scenes, Cloudflare's Wrangler [uses the `unenv` library's polyfills](https://blog.cloudflare.com/more-npm-packages-on-cloudflare-workers-combining-polyfills-and-native-code/) where natively supported Node APIs fall short. Based on the Wrangler package's source code, I've updated Partykit's `nodejs-compat-plugin` to do the same. 

## Changes made
* Set up reliable means of getting relative path mapping to root of package, as seen [here](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/paths.ts#L61) and [here](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/scripts/bundle.ts#L26).
* Update nodejs-compat plugin to combine previously supported Node.js APIs with unenv polyfills. We combine `unenv`'s `cloudflare` and `nodeless` presets. The plugin now closely resembles the one seen [here](https://github.com/cloudflare/workers-sdk/blob/main/packages/wrangler/src/deployment-bundle/esbuild-plugins/hybrid-nodejs-compat.ts). This is the plugin Wrangler uses [(source)](url).
* Clean up nodejs-compat plugin usages in `cli.tsx` and `dev.tsx` previously needed to support Node.js APIs, that are not needed anymore.

## Impact
* Hard-coded names of supported Node.js APIs are no longer present. The combination of `cloudflare` and `nodeless` presets in `unenv` are used instead. API support can be viewed [in this matrix's right-most column](https://workers-nodejs-compat-matrix.pages.dev/), and [unenv docs](https://workers-nodejs-compat-matrix.pages.dev/).
* PartyKit users can now access the same Node.js APIs that they would with Wrangler.

## Notes
As a proof-of-concept, I forked the [`partykit` NPM package](https://www.npmjs.com/package/@maryam-dev/partykit) and made these same changes (plus fork setup). Then, I created a [starter React + PartyKit project](https://github.com/maryam-khan-dev/farmer-drive), which simply shows that the `os` API now works.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21082,7 +21082,6 @@
       "version": "3.2.3",
       "resolved": "https://registry.npmjs.org/consola/-/consola-3.2.3.tgz",
       "integrity": "sha512-I5qxpzLv+sJhTVEoLYNcTW+bThDCPsit0vLNKShZx6rLtpilNpmmeTPaeqJb9ZE9dV3DGaeby6Vuhrw38WjeyQ==",
-      "dev": true,
       "engines": {
         "node": "^14.18.0 || >=16.10.0"
       }
@@ -22339,8 +22338,7 @@
     "node_modules/defu": {
       "version": "6.1.4",
       "resolved": "https://registry.npmjs.org/defu/-/defu-6.1.4.tgz",
-      "integrity": "sha512-mEQCMmwJu317oSz8CwdIOdwf3xMif1ttiM8LTufzc3g6kR+9Pe236twL8j3IYT1F7GfRgGcW6MWxzZjLIkuHIg==",
-      "dev": true
+      "integrity": "sha512-mEQCMmwJu317oSz8CwdIOdwf3xMif1ttiM8LTufzc3g6kR+9Pe236twL8j3IYT1F7GfRgGcW6MWxzZjLIkuHIg=="
     },
     "node_modules/delaunator": {
       "version": "5.0.0",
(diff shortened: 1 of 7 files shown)


threepointone:
This is still on my list of things to review, sorry for the delay

threepointone (approved):
This looks great to me, thanks for all this effort. Approving and landing.

threepointone:
This is released in partykit@beta, lemme know if it works locally and in production?

maryam-khan-dev:
> This is released in partykit@beta, lemme know if it works locally and in production?

Hey Sunil, happy new year and thank you so much for taking the time to review my PR! I tested things out carefully today, and I noticed some errors/room for improvement. This PR makes these 3 mistakes:
1. Unintentionally rolls back some compatibility due to fully relying on `unenv`'s polyfills. `unenv` still mocks some methods, so the original approach of directly exposing modules in [dev.tsx](https://github.com/partykit/partykit/blob/175a9776e6470f0d76c1e671f4700beed594dfe7/packages/partykit/src/dev.tsx#L964) and [cli.tsx](https://github.com/partykit/partykit/blob/175a9776e6470f0d76c1e671f4700beed594dfe7/packages/partykit/src/cli.tsx#L961) must be preferred when possible. 
2. Falls behind Wrangler due to using `unenv` v1.10.0, when [Wrangler uses `unenv` 2.0.0](https://github.com/cloudflare/workers-sdk/blob/6c2e94318dd079f9cba3f763e67ba6698ef597f6/packages/wrangler/package.json#L79). There is a substantial difference in polyfill support between the two. Full disclosure: `unenv` v2 is still in development, but Wrangler's using it so I think it's safe for us to use too.
3. Unintentionally rolls back support for `cloudflare:sockets` and `cloudflare:email` import paths. These must be added back in.

I made PR #934, which fixes these mistakes. Heads up - it looks like a huge PR because of the `package-lock.json` file! In reality, it is 154 lines long :)

PR #934, opened 2025-01-22
Restore & improve Node APIs' compatibility with PartyKit

maryam-khan-dev:
My earlier PR, #928, added more Node.js polyfills to PartyKit with the `unenv` library, following the Wrangler source code. However, it contains a few errors and has fallen behind Wrangler due to an older `unenv` version. 

## Summary of changes
* Switches PartyKit over to `unenv` version 2 [to match Wrangler](https://github.com/cloudflare/workers-sdk/blob/6c2e94318dd079f9cba3f763e67ba6698ef597f6/packages/wrangler/package.json#L79). So far version 1.10.0, which contains a lot less functionality, was used. Common APIs, like `node:timers`, are only mocked in v1.10.0. 
* Prefers directly injecting the APIs where possible, and adds `unenv` polyfills otherwise. Before PR # 928,  [the default supported APIs](https://developers.cloudflare.com/workers/runtime-apis/nodejs/#built-in-nodejs-runtime-apis) were 'injected' as part of build setup (see [here](https://github.com/partykit/partykit/blob/175a9776e6470f0d76c1e671f4700beed594dfe7/packages/partykit/src/dev.tsx#L964) and [here](https://github.com/partykit/partykit/blob/175a9776e6470f0d76c1e671f4700beed594dfe7/packages/partykit/src/nodejs-compat.ts#L22C5-L28C8)). Then, # 928 completely replaced that with `unenv`'s presets. After careful inspection, I noticed that this approach can reduce some compatibility (e.g. no `util.callbackify()`). So I have now reinstated directly injecting the default APIs, and added `unenv`-backed extra compatibility wherever relevant.

The only exception to this rule is `node:process`, which is a default supported API. But since its directly-injected version exposes far less functionality than `unenv`'s corresponding polyfill (no `process.cwd()`, for instance), the `unenv` polyfill is preferred for max compatibility.

* Marks `unenv` as external in the build script, to avoid errors due to `unenv/runtime` code (which must be resolved and read during bundling). This also follows [Wrangler's way of resolving the issue](https://github.com/cloudflare/workers-sdk/blob/6c2e94318dd079f9cba3f763e67ba6698ef597f6/packages/wrangler/scripts/deps.ts#L25).
* Adds support for `cloudflare:sockets` and `cloudflare:email` import paths back in, after they were accidentally removed in # 928.


maryam-khan-dev:
Seems that checks are failing due to patch-package [applying outdated patches](https://github.com/partykit/partykit/actions/runs/12900600918/job/35971334386?pr=934). Looks like they may not be needed?

threepointone:
Thanks for this. I'll fix the build errors soon

threepointone (approved):
thanks so much for this. I finally got the build to pass, needed a bunch of types being massaged etc. landing now.

Issue #953, opened 2025-05-20
Is this maintained or developed further? Enterprise on-prem deployment license


v1adko:
Hello team, I've [made a post on your discord](https://discordapp.com/channels/1051830863576453180/1372663889191374959) a few days ago, like instructed by your CLI when trying to deploy to our own cloudflare.

Is this service and library actively worked on and maintained? Is it suitable if we want to use it for production?
Want to understand the risks and what we're getting ourselves into before taking this on as a product dependency.

Happy to discuss enterprise pricing, but your discord seems dead at this time.

We really love the developer experience, and have developed a solid proof of concept but not sure if we're gonna have trouble pushing this to prod on your free plan without any admin dashboard, usage statistics, billing, etc.)

750~1000 concurrent connections (and growing) for live presence indication and notifications.

CC: @threepointone 🙏 

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/ef109176-5dc2-418b-8970-2eb0d9e9d20c" />

SarthakDh:
@v1adko Did you get any response on this? We're running into the same issue.

v1adko:
> @v1adko Did you get any response on this? We're running into the same issue.

The official approach is to move away from using partykit and use partyserver directly from new official cloudflare repository.

PR #967, opened 2025-09-11
Update README to point to cloudflare/partykit

danielcompton:
Added note about current development location. I'm not sure if you want to recommend people migrate to that version or not, but I think it would at least be helpful to point it out as the place where active development is going on.

https://github.com/partykit/partykit/issues/953, https://github.com/cloudflare/partykit/issues/234

@threepointone what do you think about adding this as a pointer?


threepointone:
That's great. Thank you!

PR #984, opened 2026-04-08
perf: migrate to tinyclip from clipboardy

florian-lefebvre:
Tinyclip is [really lighter](https://npmx.dev/compare?packages=tinyclip,clipboardy) than clipboardy. This PR was made as part of the [e18e initiative](https://e18e.dev/).


packages/partykit/scripts/build.ts
```diff
@@ -59,7 +59,6 @@ esbuild.buildSync({
     "esbuild",
     "fsevents",
     "miniflare",
-    "clipboardy",
```
florian-lefebvre, line 62: Unlike clipboardy, tinyclip can be safely inlined as it does not come with any fallback

PR #987, opened 2026-08-05
fix: set correct Content-Type for .ico (and other) static assets

### LINKED ISSUE #986: Static .ico assets served as text/plain (favicon broken in browsers/Google)
## Problem

PartyKit serves `.ico` files (e.g. `/favicon.ico`) with:

```
Content-Type: text/plain; charset=utf-8
X-Content-Type-Options: nosniff
```

Browsers and Google Search refuse to treat that as an image, so favicons show as the default globe.

## Root cause

`packages/partykit/facade/fetch-static-asset.ts` imports `mime/lite`. Lite deliberately drops all `vnd.*` / `x-*` types. `.ico` maps to the IANA type `image/vnd.microsoft.icon`, which only exists in the **full** `mime` DB — so `mime.getType('favicon.ico')` is `null` under lite, and the asset pipeline falls back to `text/plain`.

This is the same class of bug Nitro fixed by switching Cloudflare deploys to the full mime DB: https://github.com/nitrojs/nitro/pull/933

Reproduced on a PartyKit-hosted site (`lolsketch.com/favicon.ico`) before working around it in app code.

## Expected

`/favicon.ico` should be served as `image/vnd.microsoft.icon` (or `image/x-icon`).

## Suggested fix

Use `mime` (full) instead of `mime/lite` when determining Content-Type for static assets, and set `Content-Type` on the asset response from the file extension (the CDN currently returns `text/plain` for unknown types).

A PR is incoming.

devonpmack:
## Summary

- Fixes #986: `.ico` favicons were served as `text/plain` because `mime/lite` omits `vnd.*` types (`image/vnd.microsoft.icon`).
- Switch the static-asset facade to the full `mime` package and rewrite `Content-Type` on asset responses from the file extension.
- Same approach Nitro took for Cloudflare: https://github.com/nitrojs/nitro/pull/933

## Test plan

- [ ] Deploy a PartyKit project with `public/favicon.ico`
- [ ] `curl -sI https://<host>/favicon.ico` → `Content-Type: image/vnd.microsoft.icon`
- [ ] Opening the URL in a browser shows the icon (not a download/plaintext)
- [ ] Existing `image/png` / `text/css` / `application/javascript` assets still get correct types
- [ ] SPA fallback still serves `index.html` as `text/html`


devonpmack:
Hi @threepointone, let me know what you think, just a small bugfix 🙏

